PR #40, opened 2025-11-04
`mul_shr`implementation and coverage

### LINKED ISSUE #16: [Feature]: `mul_shr` function
Implement a function that calculates `floor(x * y >> n)` with full precision.

ericnordelo:
Resolves #16 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog

## Release Notes

* **New Features**
  * Introduced multiply-shift-right operation with rounding support across all integer types.
  * Enhanced multiply-divide operations with improved overflow detection and return values.
  * Added comprehensive rounding modes for arithmetic operations.

* **Tests**
  * Expanded test coverage for arithmetic operations, edge cases, and rounding behavior.

### DIFF
--- a/math/core/sources/macros.move
+++ b/math/core/sources/macros.move
@@ -262,14 +262,12 @@ public(package) fun mul_div_u256_wide(
         denominator,
     );
     if (overflow) {
-        return (true, 0)
-    };
-
-    if (remainder != 0) {
-        return round_division_result(quotient, denominator, remainder, rounding_mode)
-    };
-
-    (false, quotient)
+        (true, 0)
+    } else if (remainder == 0) {
+        (false, quotient)
+    } else {
+        round_division_result(quotient, denominator, remainder, rounding_mode)
+    }
 }
 
 /// Internal helper for `mul_shr` that selects the most efficient implementation based on the input size.
@@ -343,8 +341,8 @@ public(package) fun mul_shr_u256_wide(
     rounding_mode: RoundingMode,
 ): (bool, u256) {
     let product = u512::mul_u256(a, b);
-    let hi = u512::hi(&product);
-    let lo = u512::lo(&product);
+    let hi = product.hi();
+    let lo = product.lo();
 
     if (shift == 0) {
         if (hi != 0) {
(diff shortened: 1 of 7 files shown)


bidzyyys (approved):
Looks good!

math/core/sources/macros.move
```diff
@@ -221,38 +266,143 @@ public(package) fun mul_div_u256_wide(
...
+/// and `result` contains the shifted and rounded value when no overflow occurred.
+public(package) fun mul_shr_u256_wide(
+    a: u256,
+    b: u256,
+    shift: u8,
+    rounding_mode: RoundingMode,
+): (bool, u256) {
+    let product = u512::mul_u256(a, b);
+    let hi = u512::hi(&product);
+    let lo = u512::lo(&product);
```
immrsd:
```suggestion
    let hi = product.hi();
    let lo = product.lo();
```

math/core/sources/macros.move
```diff
@@ -221,38 +266,143 @@ public(package) fun mul_div_u256_wide(
...
-        if (should_round_up) {
-            if (quotient == std::u256::max_value!()) {
-                return (true, 0)
-            };
-            quotient = quotient + 1;
-        }
+        return round_division_result(quotient, denominator, remainder, rounding_mode)
     };
 
     (false, quotient)
```
immrsd: We might write it using if-else and go without early returns here
```
if (overflow) {
    (true, 0)
} else if (remainder == 0) {
    (false, quotient)
} else {
    round_division_result(quotient, denominator, remainder, rounding_mode)
}
```

math/core/sources/macros.move
```diff
@@ -138,8 +138,61 @@ public(package) macro fun mul_div<$Int>(
...
+///
+/// #### Aborts
+/// Does not emit custom errors, but will inherit the Move abort that occurs when `$shift` is 256 or
+/// greater.
+public(package) macro fun mul_shr<$Int>(
+    $a: $Int,
+    $b: $Int,
+    $shift: u8,
+    $rounding_mode: RoundingMode,
+): (bool, u256) {
```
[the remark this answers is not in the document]
ericnordelo: I don't think we have to/should. As long as the macros are internal to the package, the bool for overflow is more descriptive, and also more efficient gas wise than using Option (which is a struct). We should be ok with keeping option only in the public API unless it makes sense to use it semantically.

PR #42, opened 2025-11-04
Fix shift operations return values 

### LINKED ISSUE #41: [Task]: Fix shift operations return values

Currently, we always return `option::none()` when the number of bits to shift is greater than the number of bits of the operand. It makes more sense to return 0 when the operand is 0, since no non-zero value is consumed from the shift,

ericnordelo:
Resolves #41  

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bit shift operations (checked_shl, checked_shr) across integer types to properly handle zero-value inputs.

* **New Features**
  * Added mul_div function to u8 integer type with overflow detection capability.
  * Enhanced mul_div in u64 with overflow detection relative to u64 bounds.

* **Documentation**
  * Updated shift operation documentation to reflect current implementation.

* **Tests**
  * Added comprehensive test coverage for shift operations with zero inputs across all integer types.

### DIFF
--- a/math/core/tests/u32_tests.move
+++ b/math/core/tests/u32_tests.move
@@ -37,7 +37,7 @@ fun checked_shl_returns_some() {
 
 #[test]
 fun checked_shl_zero_input_returns_zero_for_overshift() {
-    assert_eq!(u32::checked_shl(0, 12), option::some(0));
+    assert_eq!(u32::checked_shl(0, 33), option::some(0));
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


math/core/sources/u128.move
```diff
@@ -11,25 +11,29 @@ public fun average(a: u128, b: u128, rounding_mode: RoundingMode): u128 {
 /// Shift the value left by the given number of bits.
 ///
 /// Returns `None` for the following cases:
-/// - the shift is greater than or equal to 128 bits.
```
immrsd: I feel we should keep this comment but clarify that none is returned when `shift` is greater than bit width for a non-zero `value`. Applies to all packages

ericnordelo: Note that the comment that we have kept (the one below) already covers the case you are describing, because shifting greater than bit width for a value not equal to zero effectively consumes a non-zero bit. That's why I considered there's no need to keep the comment.

PR #46, opened 2025-11-12
`log2` implementation and tests

### LINKED ISSUE #22: [Feature]: `log2` function
Implement function that return the log in base 2.

immrsd:
Resolves #22

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog

## Summary by CodeRabbit

* **New Features**
  * Added logarithm base-2 (log2) computation across all standard integer types with configurable rounding modes (down, up, nearest)
  * Added multiply-shift (mul_shr) macro supporting overflow detection and multiple rounding strategies

* **Tests**
  * Added comprehensive test coverage for logarithm and multiply-shift operations

### DIFF
--- a/math/core/sources/macros.move
+++ b/math/core/sources/macros.move
@@ -492,7 +492,7 @@ public(package) fun round_division_result(
 }
 
 /// Nearest-integer rounding for log2 without floats.
-/// 
+///
 /// #### Parameters
 /// - `value`: The value being tested (already cast to u256).
 /// - `floor_log`: The threshold exponent for comparison.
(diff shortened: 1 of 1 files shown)


math/core/sources/macros.move
```diff
@@ -180,41 +217,50 @@ public(package) macro fun clz<$Int>($value: $Int, $bit_width: u16): u16 {
...
+    let power_of_two = (1 as $Int) << (floor_log as u8);
+    if (value == power_of_two) {
+        return floor_log
+    };
+    if (rounding_mode == rounding::up()) {
+        return floor_log + 1
+    };
+    // Nearest: decide whether to round up or down based on midpoint (2^n × √2)
+    let threshold_exp = 2 * floor_log + 1;
+    if (log_should_round_up(value as u256, threshold_exp)) {
```
ericnordelo: The `let threshold_exp = 2 * floor_log + 1;` should be contained in the log_should_round_up function, we can pass `floor_log` as the second argument instead of `threshold_exp`.

ericnordelo: Then for the sake of clarity, let's create an inner function then, but still move the param internally, it is a matter of how easy is to understand why this works in 6 months.

PR #47, opened 2025-11-13
ref: ownership transfer wrappers

ericnordelo:
Refactors the ownable module into wrappers with different transfer policies.

This is the proposed idea, I will add full coverage after getting confirmation on the design. We need to discuss:
- Does using dof for off-chain accessibility make sense?
- Can this be made more modular? Can the directory structure and naming be improved? Is there a way of simplifying the external interface exposure?
- Is the composability granted? Can users "use" the inner cap without issues by borrowing?

### DIFF
--- a/math/core/sources/coin_utils.move
+++ b/math/core/sources/coin_utils.move
@@ -1,0 +1,203 @@
+module openzeppelin_math::coin_utils;
+
+// === Errors ===
+
+/// Value cannot be safely cast to `u64` (exceeds `std::u64::max_value!()`).
+#[error(code = 0)]
+const ESafeDowncastOverflowedInt: vector<u8> = b"Value cannot be represented as u64";
+
+/// Decimals value is invalid (must be <= 24).
+#[error(code = 1)]
+const EInvalidDecimals: vector<u8> = b"Decimals value is invalid (must be <= 24)";
+
+// === Constants ===
+
+/// Maximum decimals supported for cross-chain transfers.
+///
+/// Set to 24 to cover all known blockchain token standards with safety margin:
+/// - Ethereum/EVM chains: typically 18 decimals (ETH, most ERC-20 tokens)
+/// - Sui/Move chains: typically 6-9 decimals
+/// - Stablecoins: typically 6 decimals (USDC, USDT)
+/// - Bitcoin: effectively 8 decimals (satoshi)
+/// - Solana: typically 9 decimals
+///
+/// The value 24 provides significant headroom beyond the practical maximum
+/// of 18-19 decimals used in production systems.
+const MAX_DECIMALS: u8 = 24;
+
+// === Public Functions ===
+
+/// Downcast a `u256` balance to `u64`, handling decimal scaling.
+///
+/// This function converts token amounts between different decimal precisions,
+/// preserving economic value while fitting within `u64` constraints.
+///
+/// **IMPORTANT: When scaling down (`source_decimals` > `target_decimals`), this
+/// function TRUNCATES fractional parts rather than rounding.** For example:
+/// - 1.999 tokens → 1 token (NOT 2)
+/// - 0.999 tokens → 0 tokens (NOT 1)
+///
+/// This behavior is standard in blockchain systems to prevent inflation but
+/// means precision is permanently lost when scaling to lower decimal places.
+///
+/// # Arguments
+///
+/// * `raw_amount` - The original balance (e.g., from Ethereum with 18 decimals).
+/// * `source_decimals` - Source chain decimal places (must be <= 24).
+/// * `target_decimals` - Target decimal places (must be <= 24, typically 6-9 for Sui).
+///
+/// # Returns
+///
+/// The scaled balance as u64
+///
+/// # Aborts
+///
+/// * `EInvalidDecimals` - If either decimal value exceeds `MAX_DECIMALS` (24).
+/// * `ESafeDowncastOverflowedInt` - If scaled amount exceeds `std::u64::max_value!()`.
+///
+/// # Examples
+///
+/// ```
+/// // Scaling down: Ethereum to Sui (precision preserved for clean values)
+/// // 1.0 token with 18 decimals = 1000000000000000000
+/// // 1.0 token with 9 decimals = 1000000000
+/// let sui_amount = safe_downcast_balance(1000000000000000000, 18, 9);
+/// assert!(sui_amount == 1000000000, 0);
+///
+/// // Scaling down with truncation (fractional part lost)
+/// // 1.999999999 tokens with 9 decimals = 1999999999
+/// // Scaled to 0 decimals = 1 (NOT 2 - truncates, does not round)
+/// let truncated = safe_downcast_balance(1999999999, 9, 0);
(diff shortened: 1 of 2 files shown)


contracts/access/sources/ownership_transfer/delayed.move
```diff
@@ -0,0 +1,216 @@
...
+/// guaranteeing an on-chain lead time for sensitive capability moves.
+///
+/// Unwrapping follows the same delayed pattern, ensuring the capability cannot be reclaimed
+/// without respecting the delay period.
+module openzeppelin_access::delayed_transfer;
+
+use sui::clock;
+use sui::dynamic_object_field as dof;
+use sui::event;
+
```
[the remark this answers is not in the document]
ericnordelo: This modules are not required since they are part of the prelude.

contracts/access/tests/two_step_tests.move
```diff
@@ -0,0 +1,49 @@
...
+        sui::tx_context::dummy_tx_hash_with_hint(hint),
+        0,
+        0,
+        0,
+    )
+}
+
+fun new_cap(ctx: &mut TxContext): DummyCap {
+    DummyCap { id: sui::object::new(ctx) }
+}
```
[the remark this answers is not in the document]
ericnordelo: These modules are not required since they are part of the prelude.

contracts/access/sources/ownership_transfer/delayed.move
```diff
@@ -0,0 +1,216 @@
...
+public struct DelayedTransferWrapper<phantom T: key + store> has key {
+    id: UID,
+    min_delay_ms: u64,
+    pending: option::Option<PendingTransfer>,
+}
+
+/// Snapshot of a scheduled transfer or unwrap, including the execution time.
+/// A non-existing recipient means an unwrap is scheduled.
+public struct PendingTransfer has drop, store {
+    recipient: option::Option<address>,
```
damirka: `Option` type [is imported implicitly](https://move-book.com/move-basics/standard-library#implicit-imports).

```suggestion
    recipient: Option<address>,
```

contracts/access/sources/ownership_transfer/delayed.move
```diff
@@ -0,0 +1,216 @@
...
+        id: object::new(ctx),
+        min_delay_ms,
+        pending: option::none(),
+    };
+    dof::add(&mut wrapper.id, WRAPPED_FIELD, cap);
+    wrapper
+}
+
+/// Borrow the wrapped capability immutably—useful for inspection without touching the schedule.
+public fun borrow<T: key + store>(wrapper: &DelayedTransferWrapper<T>): &T {
```
damirka: If you want this to be supported in PTB context, you need to allow borrowing _by value_. See [borrowing items from Kiosk](https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/sui-framework/sources/kiosk/kiosk.move#L600).

damirka: Otherwise you'd force everyone using it to implement a custom function in their packages to support it. This will be changed eventually, but currently there's no other way to provide PTB support.

contracts/access/sources/ownership_transfer/delayed.move
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// === Scheduling / delay management ===
+
+/// Schedule a new transfer to `new_owner`. Stores recipient + deadline; caller later invokes
+/// `execute_transfer` after `min_delay_ms` has passed.
+public fun schedule_transfer<T: key + store>(
+    wrapper: &mut DelayedTransferWrapper<T>,
+    new_owner: address,
+    clock: &clock::Clock,
```
damirka: Import `Clock` at top level.

```suggestion
    clock: &Clock,
```

contracts/access/tests/delayed_tests.move
```diff
@@ -0,0 +1,69 @@
...
+
+    delayed_transfer::schedule_transfer(&mut wrapper, recipient, &clk, owner);
+    let scheduled = event::events_by_type<delayed_transfer::TransferScheduled>();
+    assert!(std::vector::length(&scheduled) == 1, 0);
+
+    clock::set_for_testing(&mut clk, 10);
+    delayed_transfer::execute_transfer(wrapper, &clk, &mut ctx);
+
+    let executed = event::events_by_type<delayed_transfer::OwnershipTransferred>();
+    assert!(std::vector::length(&executed) == 1, 1);
```
damirka: Don't use abort codes in tests. 
For expected results use `assert_eq!` from `std::unit_test`.

```suggestion
    use std::unit_test::assert_eq;

    assert_eq!(executed.length(), 1);
```

contracts/access/sources/ownership_transfer/two_step.move
```diff
@@ -0,0 +1,128 @@
...
+        wrapper_id,
+        new_owner: ctx.sender(),
+    };
+    event::emit(OwnershipRequested {
+        wrapper_id,
+        current_owner,
+        requester: ctx.sender(),
+    });
+    transfer::transfer(request, current_owner);
+}
```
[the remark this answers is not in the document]
ericnordelo: Good points. Couple of comments:
- Is T in your example the wrappper or the cap? Are you suggesting to wrap the cap only for transfers and not in general?
- Note that either way wrapping in PendingTransfer and sharing the object would make the capability shared, and we would need to check the address of the current owner in the borrow function. This probably make sense, but I'm trying to avoid sharing the object if possible to leverage the fast path and avoid going through consensus. In this case this may not be as important and is probably better to allow accepting directly, since it does also avoids potentially accumulating spam requests.

Curious on @damirka thoughts regarding how significant is fast path vs consensus here (my intuition is that sharing shouldn't be that problematic).

contracts/access/sources/ownership_transfer/delayed.move
```diff
@@ -0,0 +1,232 @@
...
+    self: &mut DelayedTransferWrapper<T>,
+    capability: T,
+    borrow: Borrow,
+) {
+    let Borrow { wrapper_id, object_id } = borrow;
+
+    assert!(object::id(self) == wrapper_id, EWrongDelayedTransferWrapper);
+    assert!(object::id(&capability) == object_id, EWrongDelayedTransferObject);
+
+    dof::add(&mut self.id, Wrapped {}, capability);
```
ericnordelo: I have to check BCS representation in deep for this cases, but I wonder if using an empty struct here is better than a vector with 5 a 5 bytes word for example? cc @damirka

contracts/access/sources/ownership_transfer/two_step.move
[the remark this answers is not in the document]
ericnordelo: Got it, we should then probably make it a shared object only when a transfer is pending in the two step process. This would add the overhead of having to always check the caller address when borrowing, but it would make the flow easier/similar to what we have in other ecosystems, avoiding spamming requests, among the other advantages mentioned in this thread.

bidzyyys (commented):
Looks good for the first pass

bidzyyys (approved):
LGTM!

immrsd (approved):
LGTM

PR #53, opened 2025-11-19
feat: `sqrt` implementation and tests

### LINKED ISSUE #21: [Feature]: `sqrt` function
Implement function that return the square root of a number.

immrsd:
Resolves #21

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog

## Summary by CodeRabbit

* **New Features**
  * Added square root operations for all unsigned integer types with configurable rounding modes (down, up, nearest), enabling flexible mathematical computations with precise control over result rounding.

* **Tests**
  * Comprehensive test coverage added for square root operations across all integer widths, including edge cases and boundary conditions.

### DIFF
--- a/math/core/sources/internal/common.move
+++ b/math/core/sources/internal/common.move
@@ -31,6 +31,19 @@ public(package) fun clz(val: u256, bit_width: u16): u16 {
     count
 }
 
+/// Return the position of the most significant bit (MSB) in an unsigned integer value of arbitrary bit width.
+///
+/// This function returns the zero-based index of the most significant set bit in a value of a given bit width
+/// (such as u8, u16, u32, u64, u128, or u256). The MSB position is calculated as `bit_width - 1 - clz(val, bit_width)`,
+/// where `clz` is the count of leading zeros. For a zero input, the function returns 0 by convention.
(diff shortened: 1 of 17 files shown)


ericnordelo (approved):
Great job @immrsd. 

Only comment, non-blocking and for your consideration (fill free to chose either way) is that maybe it makes sense to return the previous integer type on each public function, since it fits, and then people don't need to use try_as to convert if they need it.

immrsd:
> Only comment, non-blocking and for your consideration (fill free to chose either way) is that maybe it makes sense to return the previous integer type on each public function, since it fits, and then people don't need to use try_as to convert if they need it.

@ericnordelo I’ve considered that option but don’t think it’s feasible. The result of `sqrt` can still overflow a smaller `uint` type. For instance, `sqrt(65_535)` (max `u16`) is `255.99` and rounds to `256`, which exceeds the maximum `u8` value.

On top of that, I find it counter-intuitive to return a `uint` type different from the one used in the operation. In most cases `sqrt` will be part of a larger calculation that consistently operates on one specific `uint` type, so returning a different type would just force callers to cast the value back up anyway.

PR #104, opened 2025-11-28
`log10` implementation and tests

### LINKED ISSUE #23: [Feature]: `log10` function
Implement function that return the log in base 10.

immrsd:
Resolves #23 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog

## Release Notes

* **New Features**
  * Base-10 logarithm (log10) function now available for all unsigned integer types with configurable rounding modes (down, up, nearest).
  * Zero values return 0; exact powers of 10 return precise exponents.

* **Tests**
  * Comprehensive test coverage added for log10 across all integer types and rounding scenarios.


math/core/sources/u128.move
```diff
@@ -87,6 +87,13 @@ public fun log256(value: u128, rounding_mode: RoundingMode): u8 {
     macros::log256!(value, BIT_WIDTH as u16, rounding_mode)
 }
 
+/// Compute the log in base 10 of a positive value with configurable rounding.
+///
+/// Returns 0 if given 0.
+public fun log10(value: u128, rounding_mode: RoundingMode): u8 {
```
bidzyyys: Should it return `u8` or `u128`? Thinking about the proper attitude for each data type.

math/core/sources/internal/macros.move
```diff
@@ -340,6 +340,91 @@ public(package) macro fun log256<$Int>(
...
+/// - `$Int`: Any unsigned integer type (`u8`, `u16`, `u32`, `u64`, `u128`, or `u256`).
+///
+/// #### Parameters
+/// - `$value`: The unsigned integer to compute the logarithm for.
+/// - `$rounding_mode`: Rounding strategy drawn from `rounding::RoundingMode`.
+///
+/// #### Returns
+/// The base-10 logarithm as a `u8`, rounded according to the specified mode.
+/// Returns `0` if `$value` is 0.
+public(package) macro fun log10<$Int>($value: $Int, $rounding_mode: RoundingMode): u8 {
```
bidzyyys: Thinking if it should return `u8` or `$Int`...

math/core/sources/u128.move
immrsd: I believe it's better to return `u8` in such cases for the following reasons:
- the result is guaranteed to fit into `u8`
- some users of the lib will need `u128` for further calculations, some will need `u8`; it's easy and safe to cast up from `u8`, but casting down from `u128` may overflow and special handling is required in that case
- in some calculations `log10` result might be used as an `exponent: u8` parameter in the [std::pow](https://docs.sui.io/references/framework/sui_std/u64#std_u64_pow) function
- the behaviour is consistent throughout the library

PR #121, opened 2025-12-10
`Pausable` module

### LINKED ISSUE #108: [Feature]: Pausable base component

Pausable boilerplate

immrsd:
Resolves #108

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog

## Summary by CodeRabbit

* **New Features**
  * Introduces a pausable mechanism for managing emergency stops with pause/unpause functionality, event emissions, and guard functions to enforce pause states in operations.

* **Tests**
  * Adds comprehensive test coverage for the new pausable functionality, including state transitions and error cases; refactors test helper utilities.

### DIFF
--- a/contracts/access/tests/delayed_tests.move
+++ b/contracts/access/tests/delayed_tests.move
@@ -1,22 +1,16 @@
+#[test_only]
 module openzeppelin_access::delayed_tests;
 
 use openzeppelin_access::delayed_transfer;
+use openzeppelin_access::test_helpers::dummy_ctx_with_sender;
 use std::unit_test::assert_eq;
 use sui::clock;
 use sui::event;
 
-#[test_only]
 public struct DummyCap has key, store {
     id: sui::object::UID,
 }
 
-#[test_only]
-public fun dummy_ctx_with_sender(sender: address): TxContext {
-    let tx_hash = x"3a985da74fe225b2045c172d6bd390bd855f086e3e9d525b46bfe24511431532";
-    tx_context::new(sender, tx_hash, 0, 0, 0)
-}
-
-#[test_only]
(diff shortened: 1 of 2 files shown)


bidzyyys (commented):
Looks good to me in general!

Two questions:
1. Should it be added under `contracts/access`?
2. IMHO we should have some information about restricting access to the `Pausable` component, wdyt?

ericnordelo (commented):
In this design with only one Pausable object, I don’t think integration will work well. If you take a simple pausable transfer in a token, you need to pass the Pausable object into transfer somehow to call is_paused inside the function. That means this object needs to be shared so that users can call transfer, but if it’s shared then anyone could pause or unpause. You could wrap this object into another one, but then users would need to share the wrapper, and we’d need to expose a borrow for the is_paused logic and a borrow_mut gated by some capability for pausing/unpausing. This is the kind of pattern that should be handled in the package we provide by default, and we should have a clear, opinionated integration design for it.

Another issue is that we can’t let the argument list of functions in modules using our library grow indefinitely as we add more dependencies on standalone objects for each new feature. If users are expected to wrap things, we should design that path to be ergonomic from the start. I haven’t thought deeply enough to propose a final solution yet, but my feeling is that we could/should have something like a PackageController registry: a shared object that anyone can pass into their entry functions, plus additional capabilities to administer the controller per package. Those caps would be held by admins. The PackageController could expose a paused status, a version/status field for handling upgrades, and potentially other things like maintenance/lockdown modes, feature flags, migration metadata, or hooks for future governance controls.

I’d pass these thoughts to the Sui team to get their perspective on how they envision these integrations, but we should keep in mind that the design for modules like this needs to be considered very carefully, since breaking backward compatibility later is not straightforward.

PR #127, opened 2025-12-18
docs: add CHANGELOGs

ericnordelo:
This PR adds a CHANGELOG to each package, documenting the relevant changes locally, as well as a root-level CHANGELOG for the library that aggregates changes per release.

The CHANGELOGs follow SemVer. While only minor and patch bumps are allowed between published releases, pre-releases and release candidates may still include breaking changes, so it can be necessary to include Removed or Modified entries in those cases.

## Summary by CodeRabbit

* **New Features**
  * Two-step transfer module for capability transfers
  * Delayed transfer module with clock-based configurable delays
  * Rounding mode utilities for arithmetic operations
  * Mathematical modules with averaged computations, logarithms, and square root functions
  * 512-bit integer type for intermediate calculations
  * Coin conversion utilities for balance conversions across decimal domains

* **Documentation**
  * Updated changelogs documenting new additions


bidzyyys (approved):
Looks good to me!

PR #133, opened 2026-01-20
feat: add quick_sort and quick_sort_by implementations

### LINKED ISSUE #95: [Feature]: `quick_sort`

Follow the [reference](https://github.com/MystenLabs/deepbookv3/blob/main/packages/deepbook/sources/helper/math.move#L103-L128).

qalisander:
Casual quick sort in-place implementation similar to openzeppelin-solidity [qsort](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/239795bea728c8dca4deb6c66856dd58a6991112/contracts/utils/Arrays.sol#L115).

Compare to referenced (in issue) implementation it utilities stack data structure, since call stack (recursion) is not supported within move macro. Move macro cannot call itself.

Resolves #95 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog

## Summary by CodeRabbit

* **New Features**
  * Added in-place vector sorting functionality with new public macros for efficient sorting of unsigned integer vectors.
  * Support for multiple integer widths (u8, u16, u32, u128, u256).

* **Tests**
  * Comprehensive test suite added covering edge cases, boundary conditions, and various data patterns.

### DIFF
--- a/math/core/tests/macros/quick_sort.move
+++ b/math/core/tests/macros/quick_sort.move
@@ -525,7 +525,7 @@ fun quick_sort_by_descending_already_sorted() {
 }
 
 #[test_only]
-public struct Transfer has drop, copy {
+public struct Transfer has copy, drop {
     id: u8, // Transfer identifier
     value: u64, // Transfer value in smallest unit
 }


ericnordelo (commented):
Looking good, we also need to make this function accesible by exposing it in the corresponding integer modules.

math/core/sources/internal/macros.move
```diff
@@ -984,3 +984,71 @@ public(package) macro fun binary_search<$Int>($haystack: vector<$Int>, $needle:
...
+/// #### Parameters
+/// - `$vec`: A mutable reference to the vector to be sorted in-place.
+///
+/// #### Example
+/// ```move
+/// let mut vec = vector[3u64, 1, 4, 1, 5, 9, 2, 6];
+/// macros::quick_sort!(&mut vec);
+/// // vec is now [1, 1, 2, 3, 4, 5, 6, 9]
+/// ```
+public(package) macro fun quick_sort<$Int>($vec: &mut vector<$Int>) {
```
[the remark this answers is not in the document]
ericnordelo: Because hopefully stable_sort don't use only merge_sort, but also optimizations (like insertion for small size vectors), and also because the API is easier to understand calling stable_sort without having to understand the inner algorithm.

math/core/sources/vector.move
```diff
@@ -0,0 +1,133 @@
...
+/// let mut vec = vector[3u64, 1, 4, 1, 5, 9, 2, 6];
+/// vector::quick_sort_by!(&mut vec, |x: &u64, y: &u64| *x <= *y);
+/// // vec is now [1, 1, 2, 3, 4, 5, 6, 9]
+///
+/// // Sort in descending order
+/// let mut vec = vector[3u64, 1, 4, 1, 5, 9, 2, 6];
+/// vector::quick_sort_by!(&mut vec, |x: &u64, y: &u64| *x >= *y);
+/// // vec is now [9, 6, 5, 4, 3, 2, 1, 1]
+/// ```
+public macro fun quick_sort_by<$Int>($vec: &mut vector<$Int>, $le: |&$Int, &$Int| -> bool) {
```
ericnordelo:
```suggestion
public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> bool) {
```

This should work for any comparable type in theory

qalisander: added test cases with structs

PR #135, opened 2026-01-28
feat: add median implementation

### LINKED ISSUE #93: [Feature]: `median`

Follow the [reference](https://github.com/MystenLabs/deepbookv3/blob/main/packages/deepbook/sources/helper/math.move#L58-L74).

qalisander:
Resolves #93

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog

## Summary by CodeRabbit

* **New Features**
  * Added vector median calculation supporting configurable rounding modes for unsigned integer types (u8–u256).
  * Returns `None` for empty vectors; odd-length vectors return the middle element, even-length vectors return the rounded arithmetic mean of the two central elements.

* **Documentation**
  * Updated README with vector operations section documenting new median functionality and usage examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/contracts-sui/pull/135?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ### `openzeppelin_math`
 
+#### Added
+
+- `vector::median` macro for unsigned integer vectors (`u8`..`u256`) with configurable rounding for even-length input; returns `Option` to signal empty input. (###)
```
bidzyyys:
```suggestion
- `vector::median` macro for unsigned integer vectors (`u8`..`u256`) with configurable rounding for even-length input; returns `Option` to signal empty input. (#135)
```

math/core/sources/vector.move
```diff
@@ -194,3 +197,45 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+/// - `$Int`: Any unsigned integer type (`u8`, `u16`, `u32`, `u64`, `u128`, or `u256`).
+///
+/// #### Parameters
+/// - `$vec`: Vector whose median is desired.
+/// - `$rounding_mode`: Rounding strategy, applied only when the length is even.
+///
+/// #### Returns
+/// - `option::some(median)` when `$vec` is non-empty.
+/// - `option::none()` when `$vec` is empty — median of empty data is undefined.
+public macro fun median<$Int>($vec: vector<$Int>, $rounding_mode: RoundingMode): Option<$Int> {
```
0xNeshi: Bytecode size of inlined expansion is large, as noted in tests; consider documenting the workaround

math/core/sources/vector.move
```diff
@@ -194,3 +197,45 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+    let mut vec = $vec;
+    let len = vec.length();
+
+    // Median of empty data is undefined — signal "no value" rather than abort.
+    if (len == 0) {
+        return option::none()
+    };
+
+    quick_sort!(&mut vec);
```
0xNeshi: Sorting the vector is `O(n*log(n))`. Using quickselect would result in `O(n)` complexity for both odd and even lengths.

@bidzyyys we could consider adding a quickselect algorithm in some future milestone, as it might be useful to protocols when doing element searches. We could then update `median` to use it.

math/core/tests/macros/median.move
```diff
@@ -0,0 +1,332 @@
...
+}
+
+#[test]
+fun median_even_u256_max_max_nearest() {
+    let max = std::u256::max_value!();
+    assert_eq!(u256_median(vector[max, max], rounding::nearest()).destroy_some(), max);
+}
+
+#[test]
+fun median_even_u256_zero_and_max_down() {
```
0xNeshi, line 351: nit: for `u256`, the test suite covers `[0, MAX]` with all three rounding modes (`median_even_u256_zero_and_max_down`, `_up`, `_nearest`). For `u64`, only the `down` case is covered (`median_even_length_zero_and_max`).

Consider adding the `up` and `nearest` tests

math/core/README.md
```diff
@@ -40,3 +47,10 @@ use openzeppelin_math::{u64, rounding};
 let mean = u64::average(5, 6, rounding::down());
 // mean = 5
 ```
+
+```move
+use openzeppelin_math::{vector, rounding};
```
0xNeshi:
```suggestion
use openzeppelin_math::rounding;
use openzeppelin_math::vector;
```
nit: Move formatter produces this output

math/core/tests/macros/median.move
```diff
@@ -0,0 +1,332 @@
...
+
+#[test]
+fun median_odd_length_rounding_mode_irrelevant_up() {
+    assert_eq!(u64_median(vector[5u64, 1, 9, 3, 7], rounding::up()).destroy_some(), 5u64);
+}
+
+#[test]
+fun median_odd_length_rounding_mode_irrelevant_nearest() {
+    assert_eq!(u64_median(vector[5u64, 1, 9, 3, 7], rounding::nearest()).destroy_some(), 5u64);
+}
```
0xNeshi:
```suggestion
#[random_test]
fun median_odd_length_rounding_modes_agree_u64(mut v: vector<u64>, a: u64, b: u64) {
    // ensure `v` has odd length
    if (v.is_empty()) {
        v.push_back(a);
    } else if (v.length() % 2 == 0) {
        // pop avoids overflow risk of push if v.length == u64::max
        v.pop_back();
    };

    let down = u64_median(v, rounding::down()).destroy_some();
    let nearest = u64_median(v, rounding::nearest()).destroy_some();
    let up = u64_median(v, rounding::up()).destroy_some();

    assert_eq!(down, nearest);
    assert_eq!(nearest, up);
}
```
Consider this proptest as a more "powerful" alternative

math/core/tests/macros/median.move
```diff
@@ -0,0 +1,332 @@
...
+
+#[test]
+fun median_even_length_rounds_nearest_half_rounds_up() {
+    // (1 + 2) / 2 = 1.5 — rounding::nearest() rounds half-up
+    assert_eq!(u64_median(vector[1u64, 2], rounding::nearest()).destroy_some(), 2u64);
+}
+
+#[test]
+fun median_even_length_nearest_exact_midpoint_down() {
```
0xNeshi, line 187: Consider adding a proptest checking the result monotonicity based on rounding invariant:
```
#[random_test]
fun median_even_length_rounding_modes_are_monotonic_u64(mut v: vector<u64>, a: u64, b: u64) {
    // ensure `v` has even length (pop avoids overflow risk of push if v.length == u64::max)
    if (v.length() % 2 == 1) { v.pop_back(); };
    // ensure `v` is non-empty
    if (v.is_empty()) {
        v.push_back(a);
        v.push_back(b);
    };

    let down = u64_median(v, rounding::down()).destroy_some();
    let nearest = u64_median(v, rounding::nearest()).destroy_some();
    let up = u64_median(v, rounding::up()).destroy_some();

    assert!(down <= nearest);
    assert!(nearest <= up);
}
```

math/core/tests/macros/median.move
bidzyyys: Added!

math/core/tests/macros/median.move
bidzyyys: Added!

math/core/tests/macros/median.move
bidzyyys: Added in https://github.com/OpenZeppelin/contracts-sui/pull/135/changes/47cde156c1e402125e30f6351b7c3916c40ad864

math/core/sources/vector.move
```diff
@@ -194,3 +202,68 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+/// #### Parameters
+/// - `vec`: Vector whose median is desired (consumed).
+/// - `rounding_mode`: Rounding strategy, applied only when the length is even.
+///
+/// #### Returns
+/// - The median of `vec`.
+///
+/// #### Aborts
+/// - `EEmptyVector` if `vec` is empty.
+public(package) fun median_u256(mut vec: vector<u256>, rounding_mode: RoundingMode): u256 {
```
ericnordelo: This function should be removed for two reasons:
- It makes the public macro unusable, as the macro expands to including the internal function which is not accessible on dependant packages.
- It is used only in median, so it is not really a shared helper as the comment suggest.

math/core/sources/vector.move
```diff
@@ -194,3 +202,68 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+///
+/// #### Returns
+/// - The median of `$vec`.
+///
+/// #### Aborts
+/// - `EEmptyVector` if `$vec` is empty.
+public macro fun median<$Int>($vec: vector<$Int>, $rounding_mode: RoundingMode): $Int {
+    let vec = $vec;
+    let vec_u256 = vec.map!(|x| x as u256);
+    median_u256(vec_u256, $rounding_mode) as $Int
```
ericnordelo, line 245: We can't use internal functions inside public macros as they make them unusable.

math/core/sources/vector.move
```diff
@@ -194,3 +202,68 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+/// #### Parameters
+/// - `$vec`: Vector whose median is desired.
+/// - `$rounding_mode`: Rounding strategy, applied only when the length is even.
+///
+/// #### Returns
+/// - The median of `$vec`.
+///
+/// #### Aborts
+/// - `EEmptyVector` if `$vec` is empty.
+public macro fun median<$Int>($vec: vector<$Int>, $rounding_mode: RoundingMode): $Int {
```
ericnordelo: Do we have to consume the vector?

```suggestion
public macro fun median<$Int>($vec: &vector<$Int>, $rounding_mode: RoundingMode): $Int {
```

math/core/sources/vector.move
```diff
@@ -194,3 +202,68 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+///
+/// For odd-length vectors, the median is the middle element of the sorted vector
+/// and `$rounding_mode` has no effect on the result.
+///
+/// For even-length vectors, the median is the arithmetic mean of the two central
+/// values, rounded according to `$rounding_mode`.
+///
+/// The macro is a thin wrapper that upcasts the input to `vector<u256>` and delegates
+/// to `median_u256`, so the sort and average expansions are paid once inside the
+/// package rather than at every integrator call site.
```
ericnordelo: This is a public doc comment leaking an internal helper which is an antipattern for doc comments (we should probably add it to the skills). We should describe behavior, not the indirection in these cases.

math/core/sources/vector.move
ericnordelo: I vote we use quickselect for this implementation, even if is internal or embedded, and then we can expose it in the public API in future milestones, but mediam would be efficient. Actually, I think the real best solution and the one we should provide is benchmarking to see if sorting may be more efficient for smaller vectors, and select one algorithm or the other depending on the length.

math/core/sources/vector.move
0xNeshi: Unresolving to check if the docs are now outdated due to https://github.com/OpenZeppelin/contracts-sui/pull/325

math/core/tests/macros/median.move
```diff
@@ -1,306 +1,404 @@
...
 use openzeppelin_math::rounding::{Self, RoundingMode};
 use openzeppelin_math::vector;
 use std::unit_test::assert_eq;
 
-// Helpers for the two workhorse widths. Each test that calls a helper avoids inlining
-// `median!`'s body, keeping the test module's total bytecode under Sui's per-package
-// arena limit — every `median!` expansion includes the full `quick_sort_by!` stack
-// loop plus `average!`'s u256-upcast machinery. Widths used by only a couple of tests
-// (`u8`, `u16`, `u32`, `u128`) inline the macro directly.
+fun sorted_median_u64(sorted: &vector<u64>, rounding_mode: RoundingMode): u64 {
```
0xNeshi, line 9: This is used in a single test case, consider moving the function above the test. Feel free to ignore the suggestion.

ericnordelo (approved):
LGTM!

math/core/sources/vector.move
```diff
@@ -194,3 +202,197 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+///
+/// #### Returns
+/// - The median of `$vec`.
+///
+/// #### Aborts
+/// - `EEmptyVector` if `$vec` is empty.
+///
+/// #### Example
+/// ```move
+/// use openzeppelin_math::{rounding, vector};
```
0xNeshi:
```suggestion
/// use openzeppelin_math::rounding;
/// use openzeppelin_math::vector;
```
nit: Move formatter would output this. Same in the other example.

math/core/sources/vector.move
```diff
@@ -1,5 +1,13 @@
 module openzeppelin_math::vector;
 
+use openzeppelin_math::macros;
+use openzeppelin_math::rounding::RoundingMode;
+
+// === Errors ===
+
+#[error(code = 0)]
+const EEmptyVector: vector<u8> = "Median of empty vector is undefined";
```
immrsd: We might want to rename this constant since it relates to median function, not the whole module. If we add a new function that also restricts empty vectors, we'll have a different error code and thus will have to somehow differentiate between the two

math/core/sources/vector.move
```diff
@@ -194,3 +202,197 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+    let mut end = vec.length();
+
+    loop {
+        if (start + 1 >= end) return vec[k];
+
+        // Small ranges are cheaper to sort directly than to keep partitioning.
+        if (end - start <= 10) {
+            insertion_sort_u256_range(vec, start, end);
+            return vec[k]
+        };
```
ericnordelo: We should apply this idea to quicksort too after proper benchmarking. cc @bidzyyys

math/core/sources/vector.move
```diff
@@ -194,3 +202,197 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+            start = eq_end;
+        } else {
+            return vec[k]
+        };
+    }
+}
+
+// Sort a half-open sub-range [start, end). Used as the quickselect base case
+// for small active ranges.
+fun insertion_sort_u256_range(vec: &mut vector<u256>, start: u64, end: u64) {
```
0xNeshi, line 372: Can we use this in [quick_sort_by](https://github.com/OpenZeppelin/contracts-sui/blob/acb2c0d438c070a9f4c9101a07c51dd09c0df758/math/core/sources/vector.move#L101)?

math/core/sources/vector.move
```diff
@@ -194,3 +202,197 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+fun select_k_u256(vec: &mut vector<u256>, k: u64): u256 {
+    let mut start = 0;
+    let mut end = vec.length();
+
+    loop {
+        if (start + 1 >= end) return vec[k];
+
+        // Small ranges are cheaper to sort directly than to keep partitioning.
+        if (end - start <= 10) {
+            insertion_sort_u256_range(vec, start, end);
```
0xNeshi: Double-checking, did you benchmark this claim?

bidzyyys (approved):
LGTM

immrsd (approved):
LGTM, the only concern I raised is the name of `EEmptyVector` constant

math/core/sources/vector.move
ericnordelo: Probably, but it should be a separate issue: https://github.com/OpenZeppelin/contracts-sui/pull/135#discussion_r3342094210

math/core/sources/vector.move
ericnordelo: Yes, but the claim is too strong still, since it is probabilistically better but not for all the use cases. There are cases where is not better. I will update the comment but keep the heuristic at least for the audit.

PR #151, opened 2026-02-12
fix(H-01): reminder issue in div_rem

ericnordelo:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced division remainder handling to correctly preserve remainder calculations when quotient overflow occurs, improving accuracy of edge-case arithmetic operations.

* **Tests**
  * Added test coverage for modular multiplication with maximum-value operands.
  * Added test coverage for division operations with quotient overflow.
  * Added verification for remainder preservation during overflow scenarios.


math/core/Move.lock
bidzyyys: @ericnordelo why this change happened?

bidzyyys (approved):
Looks good to me!

math/core/sources/u512.move
```diff
@@ -141,7 +146,7 @@ public fun div_rem_u256(numerator: U512, divisor: u256): (bool, u256, u256) {
     };
 
     assert!(remainder.hi == 0, EInvalidRemainder);
-    (false, quotient, remainder.lo)
+    if (overflow) (true, 0, remainder.lo) else (false, quotient, remainder.lo)
```
0xNeshi:
```suggestion
    (overflow, quotient, remainder.lo)
```
this works if on [line 134](https://github.com/OpenZeppelin/contracts-sui/pull/151/changes#diff-44f2bf21eae330fa655b56111f411fbd5e1f896408420c1ad34b8d3a30a0c0edR134), right after `overflow` is set to `true`, the `quotient` is set to `0` (safe since once overflow occurs, we no longer calculate the `quotient`).

immrsd: It does work indeed, but frankly I don't see any benefit from it. On the other hand, the logic becomes more fragile and it'll be easier to bring a bug modifying the code. I vote we keep returning explicit zero value for `quotient` in case of overflow

ericnordelo:
> Tentative approval - should CHANGELOG be updated?

Great point. At the time of releasing we hadn't added a Changelog to the project yet, but now that we have it I added the corresponding entries on this PR. We should update from others too to have everything in the CHANGELOG for the next rc/release.

0xNeshi (approved):

PR #171, opened 2026-02-18
ci: Add LLM Security Scan Prompt

0xNeshi:
## WARNING

- **MOST OF THIS WAS CREATED USING LLM AND IS NOT YET TESTED**
- This is an _untested_ draft proposal created as a result of recent meeting with auditors who suggested we use similar prompts during development, the idea being to gauge whether this is something we even want to have in our code base
- Before this is merged, we should double- and triple-check that it is safe to do so (e.g. that it CANNOT POSSIBLY expose any sensitive API keys)

## Proposal

How this is imagined to behave:
- runs whenever `main` is updated (alternative is to run every X days/weeks to accumulate enough meaningful changes)
- analyzes the code for likely security vulnerabilities
- if any potential vulnerabilities are found, the CI job fails and we get a notification on Github
- we verify whether this is an actual vulnerability or an LLM hallucination
   - if the former - we address it
   - if the latter - we ignore it

### 1. We already have a tool that does exactly this, it's called X.

Awesome, let's integrate that instead, and close this PR!

### 2. The CI job implementation actually has problem with X, it should do Y instead.

It's excellent that you noticed this problem and suggested an improvement! The better we make the CI job now, the more bugs we'll catch when it runs later.

### 3. The CI job turns out to have too many false positives OR reports too few basic vulnerabilities that are later surfaced in the actual audit.

We try to improve the prompt to improve the job's effectiveness. In the extreme case, we determine the CI job is more often than not useless, so we remove it, and call it a day.

### 4. The risk of leaking sensitive data is too great to allow running such a CI job.

Fair critique and very important! If possible, let's try to make the LLM sandboxed enough that we feel at ease. If not possible, or we feel to uncomfortable, we shouldn't just close the PR without merging and that's that.


0xNeshi:
We'll revisit a similar workflow later.

PR #183, opened 2026-02-20
ref: Two Step Transfer (and Delayed) improvements

ericnordelo:
## Summary by CodeRabbit

# Release Notes

* **Breaking Changes**
  * Object transfer flow redesigned with updated method signatures and event types
  * Transfer operations now require additional transaction context parameter
  * Phantom type parameters added to events for type-specific indexing

* **New Features**
  * Object borrowing mechanism added to enable access to wrapped objects during pending transfers
  * New request-borrow pattern for ensuring proper object handling

* **Documentation**
  * Documentation updated to reflect object-centric ownership management terminology and workflows

### DIFF
--- a/contracts/access/sources/ownership_transfer/two_step.move
+++ b/contracts/access/sources/ownership_transfer/two_step.move
@@ -191,9 +191,6 @@ public fun borrow_val<T: key + store>(self: &mut TwoStepTransferWrapper<T>): (T,
 /// - `obj`: Object being returned.
 /// - `borrow`: Hot potato produced by `borrow_val`.
 ///
-/// #### Returns
-/// - Nothing.
-///
 /// #### Aborts
 /// - `EWrongTwoStepTransferWrapper` if `borrow` does not correspond to `self`.
 /// - `EWrongTwoStepTransferObject` if `obj` does not match the borrowed object.
@@ -248,9 +245,6 @@ public fun unwrap<T: key + store>(self: TwoStepTransferWrapper<T>, ctx: &mut TxC
 /// - `self`: Wrapper being transferred.
 /// - `new_owner`: Prospective owner who may accept the transfer.
 /// - `ctx`: Transaction context.
-///
-/// #### Returns
-/// - Nothing.
 public fun initiate_transfer<T: key + store>(
     self: TwoStepTransferWrapper<T>,
     new_owner: address,
@@ -281,9 +275,6 @@ public fun initiate_transfer<T: key + store>(
 /// - `request`: Pending transfer object created by `initiate_transfer`.
 /// - `wrapper_ticket`: TTO receiving ticket for the wrapper.
 /// - `ctx`: Transaction context.
-///
-/// #### Returns
-/// - Nothing.
 ///
 /// #### Aborts
 /// - `ENotNewOwner` if caller is not the designated new owner.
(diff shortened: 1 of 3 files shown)


contracts/access/sources/ownership_transfer/two_step.move
```diff
@@ -1,215 +1,319 @@
...
     id: UID,
     wrapper_id: ID,
+    from: address,
     new_owner: address,
 }
 
-/// Hot potato to ensure a wrapped object was returned after being taken using
-/// the `borrow_val` call.
+/// Hot potato used to ensure a wrapped object was returned after being taken using `borrow_val`.
 public struct Borrow { wrapper_id: ID, object_id: ID }
```
0xNeshi: There is a case to be made for adding a phantom type to `Borrow` as well, which is that it ensures the correct `Borrow<T>` is returned with the object `T`.

Imagine a case where two objects with different types are wrapped, `borrow_val` is called on both, and the dev is trying to `return_val` them, but they mix up the `borrow_token`s.

Currently the below test fails only at runtime, because there's no compile-time type checking on the `Borrow` type:
```move
#[test_only]
fun new_cap(ctx: &mut TxContext): DummyCap {
  DummyCap { id: object::new(ctx) }
}

#[test_only]
fun new_cap_another(ctx: &mut TxContext): AnotherDummyCap {
  AnotherDummyCap { id: object::new(ctx) }
}

#[test]
fun borrow_untyped() {
  // Borrow, mutate, and return the capability through all borrow APIs.
  let owner = @0x11;
  let mut ctx = dummy_ctx_with_sender(owner);
  let mut dc_wrapper = delayed_transfer::wrap(new_cap(&mut ctx), 5, &mut ctx); // DummyCap
  let mut adc_wrapper = delayed_transfer::wrap(new_cap_another(&mut ctx), 5, &mut ctx); // AnotherDummyCap
  let mut clk = clock::create_for_testing(&mut ctx);
  clk.set_for_testing(0);

  let (dummy_cap, dc_borrow_token) = dc_wrapper.borrow_val();
  let (another_dummy_cap, adc_borrow_token) = adc_wrapper.borrow_val();

  dc_wrapper.return_val(dummy_cap, adc_borrow_token); // oops, `AnotherDummyCap`'s borrow_token
  adc_wrapper.return_val(another_dummy_cap, dc_borrow_token); // oops, `DummyCap`'s borrow_token

  destroy(dc_wrapper);
  destroy(adc_wrapper);
  destroy(clk);
}
```

**But if the borrowing type signature holds the actual object type `Borrow<phantom T>`, then _the above test fails at compile-time_.** 

NOTE: internal `return_val` runtime asserts are still valuable in case that borrow_tokens of the same type don't get mixed up.

contracts/access/sources/ownership_transfer/two_step.move
```diff
@@ -1,215 +1,319 @@
...
+    obj
 }
 
 // === Transfer flow ===
 
-/// Create an ownership request for the wrapper and send it to the current owner. The caller becomes
-/// the prospective owner and must wait for the current owner to call `transfer`.
-public fun request<T: key + store>(wrapper_id: ID, current_owner: address, ctx: &mut TxContext) {
+/// Initiate an ownership transfer by creating a shared request and sending the wrapper to it via
+/// TTO. The caller is the current owner and `new_owner` becomes the prospective owner.
```
0xNeshi: Address typo warning, perhaps by whitelisting "TTO"
https://github.com/OpenZeppelin/contracts-sui/actions/runs/22225505507/job/64291035225?pr=183

PR #188, opened 2026-02-23
fix(N-05): Code and Documentation Inconsistencies

ericnordelo:
## Summary by CodeRabbit

* **New Features**
  * Added utility functions for mathematical operations including average calculation, bit position detection, and modular arithmetic across supported integer types.
  * Enhanced shift operations with safety bounds checking.

* **Bug Fixes**
  * Improved error handling with more specific error codes for transfer state validation and object type mismatches.

* **Documentation**
  * Standardized documentation formatting guidelines for contributors.
  * Expanded function documentation across math utilities with detailed parameter and return value descriptions.

### DIFF
--- a/contracts/access/sources/ownership_transfer/two_step.move
+++ b/contracts/access/sources/ownership_transfer/two_step.move
@@ -16,10 +16,7 @@ module openzeppelin_access::two_step_transfer;
 use sui::dynamic_object_field as dof;
 use sui::event;
 
-/// Zero-sized key used to reference the dynamically stored wrapped capability or object.
-///
-/// This marker type is never instantiated with data; it only serves as the unique key for the
-/// dynamic object field that holds the underlying capability inside `TwoStepTransferWrapper`.
+/// Dynamic field key for a wrapped object.
 public struct WrappedKey() has copy, drop, store;
 
 /// Transfer request does not correspond to the provided wrapper


CONTRIBUTING.md
```diff
@@ -25,6 +25,32 @@ Your contribution must meet these requirements:
...
+
+```move
+/// Compute something.
+///
+/// #### Parameters
+/// - `value`: Input value.
+/// - `rounding_mode`: Rounding strategy.
+///
+/// #### Returns
+/// - Rounded output value.
```
0xNeshi: Another example of format inconsistency

contracts/access/sources/ownership_transfer/delayed.move
```diff
@@ -17,16 +17,22 @@ use sui::event;
 public struct WrappedKey() has copy, drop, store;
 
 #[error(code = 0)]
+/// A transfer or unwrap is already scheduled and must be executed or cancelled first.
 const ETransferAlreadyScheduled: vector<u8> = b"Transfer already scheduled.";
```
0xNeshi:
```suggestion
/// A transfer or unwrap is already scheduled and must be executed or cancelled first.
#[error(code = 0)]
const ETransferAlreadyScheduled: vector<u8> = b"Transfer already scheduled.";
```
nit: maybe we should use the `documentation - attributes - code` convention, generally used in Rust-based languages. Once the number of doc lines increases, it will become harder to connect the attributes and the related code.

math/core/sources/internal/macros.move
```diff
@@ -5,15 +5,25 @@ use openzeppelin_math::rounding::{Self, RoundingMode};
 use openzeppelin_math::u512;
 
 #[error(code = 0)]
-const EDivideByZero: vector<u8> = b"Divisor must be non-zero";
+const EDivideByZero: vector<u8> = b"Divisor must be non-zero.";
 #[error(code = 1)]
 const EZeroModulus: vector<u8> = b"Modulus must be non-zero.";
```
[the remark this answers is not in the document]
ericnordelo: I think we can be a bit flexible with this, no need to add doc comments to every error unless we find is required for doc generation later or something (separate issue).  The delayed module is a bit more complex and errors, even while simple, are not self explanatory, that's why they have comments, but those could be removed as well.

PR #264, opened 2026-03-27
feat: improve fp math conversion API

### LINKED ISSUE #267: [M-01]: Missing Scaled Conversion Functions Between Integer and Fixed-Point Types

### LINKED ISSUE #279: [N-05]: Inconsistent SCALE Constant Type Between SD29x9 Modules

ericnordelo:
Resolves #267.

## Summary by CodeRabbit

* **New Features**
  * Added whole-number conversion functions for UD30x9 and SD29x9 with scaling and truncation variants
  * Added fallible (`try_`) and non-fallible conversion options with overflow handling
  * Added raw casting support for SD29x9

* **Documentation**
  * Clarified distinction between raw casting (bit-preserving) and semantic conversions (scale-applying)

* **Tests**
  * Added comprehensive test coverage for UD30x9 and SD29x9 conversions

* **Refactor**
  * Consolidated fixed-point constants and utilities into a shared module to reduce duplication


math/fixed_point/sources/conversions/sd29x9.move
```diff
@@ -0,0 +1,181 @@
...
+///
+/// #### Returns
+/// - The whole-number portion of `x`.
+///
+/// #### Aborts
+/// - Aborts if `x` is negative.
+public fun to_u128_trunc(x: SD29x9): u128 {
+    let (whole, is_negative) = to_parts_trunc(x);
+    assert!(!is_negative, ENegativeValue);
+    whole
```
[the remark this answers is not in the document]
bidzyyys: @ericnordelo how about this finding?

bidzyyys (approved):
Looks good to me!

math/fixed_point/sources/casting/u128.move
```diff
@@ -1,15 +1,40 @@
...
+/// Casts raw `u128` fixed-point bits into `UD30x9`.
 ///
 /// #### Parameters
-/// - `x`: Raw `u128` value to convert.
+/// - `x`: Raw fixed-point bits to wrap.
 ///
 /// #### Returns
-/// - The `UD30x9` representation of `x`.
+/// - The `UD30x9` value with raw bits `x`.
 public fun into_UD30x9(x: u128): UD30x9 {
```
0xNeshi: I may be missing some context, but why is this function even necessary when we already have `ud30x9::wrap`? Seems like API duplication, a bit confusing

math/fixed_point/sources/internal/common.move
```diff
@@ -0,0 +1,93 @@
+/// Shared helpers for fixed-point package-wide constants and conversions.
+///
+/// The public `ud30x9` and `sd29x9` modules intentionally expose low-level
+/// `wrap`/`unwrap` APIs over raw scaled representations. Conversion helpers
+/// live separately and reuse the constants in this module to keep scale-aware
+/// bounds, sign handling, and terminology consistent across the package.
+module openzeppelin_fp_math::fixed_point_common;
```
0xNeshi:
```suggestion
module openzeppelin_fp_math::common;
```
nit: verbose

the "fixed point" part is implicit in the name of the whole package, and there's no conflict with regular math's `common` module

math/fixed_point/tests/sd29x9_tests/wrap_tests.move
```diff
@@ -46,6 +47,16 @@ fun wrap_negative_one_is_all_ones() {
     assert_eq!(sd29x9::wrap(1, true).unwrap(), ALL_ONES);
 }
 
+#[test]
+fun raw_casting_from_u128_matches_wrap() {
```
0xNeshi: Perfect candidate for [random_test](https://move-book.com/testing/random-test)

math/fixed_point/sources/conversions/sd29x9.move
```diff
@@ -0,0 +1,181 @@
...
+/// `SD29x9` values back into whole integers by truncating the fractional part.
+///
+/// Move does not offer a native signed integer type suitable for this package,
+/// so conversions into `SD29x9` accept an unsigned magnitude plus a sign flag,
+/// and conversions back to a simple signed form return `(magnitude, is_negative)`.
+///
+/// This module is intentionally separate from raw casting helpers like
+/// `sd29x9::wrap`, `sd29x9::from_bits`, and `casting_u128::into_SD29x9`, which
+/// preserve raw scaled bits without applying or removing the `10^9` scale.
+module openzeppelin_fp_math::sd29x9_convert;
```
0xNeshi, line 14: Why not move to `sources/sd29x9/convert.move`?
Additionally, would align nicely with the module name `sd29x9_convert`.

Similar for UD30x9, why not move to `sources/sd30x9/convert.move`?

math/fixed_point/sources/conversions/sd29x9.move
```diff
@@ -0,0 +1,181 @@
...
+/// sign flag.
+///
+/// #### Parameters
+/// - `x`: Fixed-point value to convert.
+///
+/// #### Returns
+/// - `(magnitude, is_negative)` where `magnitude` is the truncated whole-number
+///   portion of the absolute value of `x`.
+///
+/// The sign flag is always `false` when the result is zero.
```
0xNeshi:
```suggestion
/// The sign flag is always `false` when the `magnitude` part is zero.
```
optional suggestion

math/fixed_point/sources/conversions/sd29x9.move
```diff
@@ -0,0 +1,181 @@
...
+
+/// Converts an `SD29x9` value into a truncated whole-number magnitude plus its
+/// sign flag.
+///
+/// #### Parameters
+/// - `x`: Fixed-point value to convert.
+///
+/// #### Returns
+/// - `(magnitude, is_negative)` where `magnitude` is the truncated whole-number
+///   portion of the absolute value of `x`.
```
0xNeshi, line 93:
```suggestion
/// - `(magnitude, is_negative)` where `magnitude` is the truncated integer part of `abs(x)`.
```
optional suggestion, feel free to resolve without applying

math/fixed_point/sources/conversions/sd29x9.move
```diff
@@ -0,0 +1,181 @@
...
+/// Tries to convert a non-negative `SD29x9` value into a whole `u64` by
+/// truncating its fractional part.
+///
+/// #### Parameters
+/// - `x`: Fixed-point value to convert.
+///
+/// #### Returns
+/// - `some(u64)` when `x` is non-negative and the truncated whole-number
+///   portion fits in `u64`,
+/// - `none` otherwise.
```
0xNeshi:
```suggestion
/// - `some(u64)` when `x` is non-negative and the truncated whole-number
///   portion fits in `u64`, otherwise `none`.
```
align with other `try_X` functions, e.g. [try_to_u128_trunc](https://github.com/OpenZeppelin/contracts-sui/pull/264/changes#diff-e1aaa41be6bb676e54472e55de55297ccd3ba3044cc2f8838877ed17d806a0b2R131)

math/fixed_point/sources/internal/common.move
```diff
@@ -0,0 +1,93 @@
...
+
+/// Decimal scale used by all fixed-point types in this package.
+const SCALE: u128 = 1_000_000_000; // 10^9
+
+/// All-ones mask for a 128-bit unsigned integer.
+const U128_MAX_VALUE: u128 = 0xFFFF_FFFF_FFFF_FFFF_FFFF_FFFF_FFFF_FFFF; // 2^128 - 1
```
0xNeshi: potentially unnecessary, standard library already provides this value, see [std::u128::max_value!](https://github.com/MystenLabs/sui/blob/8812c22233b30d8f48178af5e9d7ff2b2aab8f76/crates/sui-framework/packages/move-stdlib/sources/u128.move#L166)

math/fixed_point/sources/internal/common.move
```diff
@@ -0,0 +1,93 @@
...
+/// - The `10^9` scale factor promoted to `u256`.
+public(package) fun scale_u256(): u256 {
+    SCALE as u256
+}
+
+/// Returns the largest 128-bit unsigned integer value.
+///
+/// #### Returns
+/// - `2^128 - 1`.
+public(package) fun u128_max(): u128 {
```
0xNeshi: related to previous comment

math/fixed_point/sources/conversions/ud30x9.move
```diff
@@ -0,0 +1,116 @@
...
+
+/// Tries to convert a `UD30x9` value into a whole `u64` by truncating its
+/// fractional part.
+///
+/// #### Parameters
+/// - `x`: Fixed-point value to convert.
+///
+/// #### Returns
+/// - `some(u64)` when the truncated whole-number portion fits in `u64`,
+/// - `none` otherwise.
```
0xNeshi:
```suggestion
/// - `some(u64)` when the truncated whole-number portion fits in `u64`, otherwise `none`.
```

math/fixed_point/sources/conversions/sd29x9.move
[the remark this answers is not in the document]
0xNeshi: This actually seems like a valid finding. Allowing `is_negative == true` for zero whole parts would solve this issue. Do you think doing this might cause some unrelated problems?

math/fixed_point/sources/casting/u128.move
ericnordelo: `wrap` is low level API even when it can be accessed, meant mostly for internal wrapping and operations, while `into_` is higher level API intended for users of the library.

ericnordelo: `into_` is semantically more expressive, and the idea is to follow the module call pattern with these casting functions, where you would do `x.into_` to get the type you need.

With this said, I'm not sure if we should keep this module or not, since this pattern requires using private function associated (`use fun`) for standard library integer types, which is (from the POV of Damir at least) an antipattern.

ericnordelo: That's why we haven't added much to this casting module, and we should IMO focus on conversion.

math/fixed_point/sources/conversions/sd29x9.move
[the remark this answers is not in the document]
ericnordelo: Valid finding. The semantics of to_parts_trunc is ok, I woulnd't modify it. I fixed the issue by properly handling the sign in to_u128_trunc and try_to_u128_trunc.

math/fixed_point/sources/conversions/sd29x9.move
ericnordelo: Great idea, but I would rather call it convertions instead of convert. Even when we are trying to do it as much as possible, the rule of naming the file after the module name is not a rule, since modules can't be nested but files can for organization.

math/fixed_point/sources/internal/common.move
ericnordelo: I see no harm on keeping it for consistency in constants usage, and it also avoids us calling an std function

bidzyyys (approved):

math/fixed_point/sources/casting/u128.move
0xNeshi: > `wrap` is low level API even when it can be accessed, meant mostly for internal wrapping and operations, while `into_` is higher level API intended for users of the library.

If it's (from now on) meant for internal use, shouldn't it be marked with `public(package)`?

> With this said, I'm not sure if we should keep this module or not...

You're wondering whether it makes sense to move these `x.into_` functions into `sd29x9.move` and `ud30x9.move` modules, and removing the `casting/u128.move` module completely? If so, I'd be in favor of that.

math/fixed_point/sources/internal/common.move
0xNeshi: Does calling an `std` function incur additional memory/gas costs?

0xNeshi: I'd still argue this constant is unnecessary for two reasons:
1. The [sd29x9_conversions](https://github.com/OpenZeppelin/contracts-sui/pull/264/changes#diff-158d539551a724a1669528df264aecb94b41591f8626426fbc9f024a78c02963R96) module already relies on a similar macro `std::u64::max_value!()`
2. Since macros are expanded at compile-time, and this [std::u128::max_value!](https://github.com/MystenLabs/sui/blob/8812c22233b30d8f48178af5e9d7ff2b2aab8f76/crates/sui-framework/packages/move-stdlib/sources/u128.move#L166) evaluates to a constant, there's no runtime harm in invoking them in [max_ud30x9_whole](https://github.com/OpenZeppelin/contracts-sui/pull/264/changes#diff-ef2fba68250988f89e899586eacea6403896241c031f6c72327e121f1bc7a86cR77) and using them in place of all [u128_max](https://github.com/OpenZeppelin/contracts-sui/pull/264/changes#diff-ef2fba68250988f89e899586eacea6403896241c031f6c72327e121f1bc7a86cR44) invocations

0xNeshi (commented):
Left responses to previous comments (see above)

math/fixed_point/sources/internal/common.move
0xNeshi: FYI if you still think it's useful to have the local constant, I'm ok with that; feel free to resolve this thread

math/fixed_point/sources/ud30x9/ud30x9_base.move
```diff
@@ -317,12 +312,12 @@ public fun pow(x: UD30x9, exp: u8): UD30x9 {
...
     };
-    let max_value = U128_MAX_VALUE as u256;
+    let max_value = common::u128_max() as u256;
     let base = x.unwrap() as u256;
     let mut result = base;
     let times = exp - 1;
     times.do!(|_| {
-        result = result * base / SCALE_U256;
+        result = result * base / common::scale_u256();
         assert!(result <= max_value, EOverflow);
```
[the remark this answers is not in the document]
0xNeshi: Handled in https://github.com/OpenZeppelin/contracts-sui/pull/280

math/fixed_point/sources/casting/u128.move
ericnordelo: Let's then remove the casting module completly. Let's keep wrap public as a low level casting alternative to conversions.

math/fixed_point/sources/internal/common.move
ericnordelo: You're right. There's no runtime/gas benefit here since std::u128::max_value! is a compile-time macro expansion, not a runtime function call. I kept the local constant only for centralization, but since we already use std max-value macros elsewhere, duplicating u128::MAX here is probably not worth it.

Updated it.

math/fixed_point/sources/internal/common.move
```diff
@@ -0,0 +1,82 @@
...
+    MIN_SD29X9_VALUE
+}
+
+/// Returns the largest whole unsigned integer that can be converted into
+/// `UD30x9` without overflowing after scaling by `10^9`.
+///
+/// #### Returns
+/// - `floor(u128::MAX / 10^9)`.
+public(package) fun max_ud30x9_whole(): u128 {
+    (std::u128::max_value!()) / SCALE
```
0xNeshi:
```suggestion
    std::u128::max_value!() / SCALE
```
nit

math/fixed_point/sources/sd29x9/sd29x9.move
```diff
@@ -153,9 +162,9 @@ public fun unwrap(x: SD29x9): u128 {
 /// #### Returns
 /// - The two's complement of `bits`.
 public(package) fun two_complement(bits: u128): u128 {
-    let inverted = bits ^ U128_MAX_VALUE;
+    let inverted = bits ^ std::u128::max_value!();
     let sum = (inverted as u256) + 1;
-    (sum & (U128_MAX_VALUE as u256)) as u128
+    (sum & ((std::u128::max_value!()) as u256)) as u128
```
0xNeshi:
```suggestion
    (sum & (std::u128::max_value!() as u256)) as u128
```
nit: there are a couple of other places in the PR where redundant brackets are used like `(std::u128::max_value!())`

math/fixed_point/sources/sd29x9/sd29x9_base.move
```diff
@@ -426,19 +424,19 @@ public fun rshift(x: SD29x9, bits: u8): SD29x9 {
...
     };
 
     let raw = x.unwrap();
-    if ((raw & SIGN_BIT) == 0) {
+    if ((raw & common::sign_bit()) == 0) {
         from_bits(raw >> bits)
     } else {
         let shifted = raw >> bits;
-        let mask = U128_MAX_VALUE << (128 - bits);
+        let mask = (std::u128::max_value!()) << (128 - bits);
```
0xNeshi:
```suggestion
        let mask = std::u128::max_value!() << (128 - bits);
```

0xNeshi (commented):
nits

math/fixed_point/sources/internal/common.move
```diff
@@ -0,0 +1,82 @@
...
+/// - The `10^9` scale factor used to encode one whole unit.
+public(package) fun scale(): u128 {
+    SCALE
+}
+
+/// Returns the raw fixed-point scale as `u256`.
+///
+/// #### Returns
+/// - The `10^9` scale factor promoted to `u256`.
+public(package) fun scale_u256(): u256 {
```
0xNeshi: What do you think about making all of these getters into macros, so they expand into the constants, thus removing indirection (from `call site -> constant getter -> constant` into `call site -> constant`)?

math/fixed_point/sources/internal/common.move
```diff
@@ -0,0 +1,70 @@
...
+    0x8000_0000_0000_0000_0000_0000_0000_0000 // -2^127 in two's complement
+}
+
+/// Returns the largest whole unsigned integer that can be converted into
+/// `UD30x9` without overflowing after scaling by `10^9`.
+///
+/// #### Returns
+/// - `floor(u128::MAX / 10^9)`.
+public(package) macro fun max_ud30x9_whole(): u128 {
+    340_282_366_920_938_463_463_374_607_431 // floor(u128::MAX / 10^9)
```
0xNeshi, line 55: Consider using the explicit expression in code (the commented out part) instead of the hardcoded value. It's less error-prone, and Move compiler is smart enough to evaluate the constant expression and save the result in the byte code.

math/fixed_point/sources/internal/common.move
```diff
@@ -0,0 +1,70 @@
...
+///
+/// This bound applies to both positive and negative whole numbers. Because
+/// `SD29x9` stores values in signed two's-complement form, negative whole
+/// conversions accept a magnitude plus a sign flag instead of a native signed
+/// integer input.
+///
+/// #### Returns
+/// - `floor((2^127 - 1) / 10^9)`.
+public(package) macro fun max_sd29x9_whole(): u128 {
+    170_141_183_460_469_231_731_687_303_715 // floor((2^127 - 1) / 10^9)
```
0xNeshi, line 69: Same as [above](https://github.com/OpenZeppelin/contracts-sui/pull/264#discussion_r3051195834)

bidzyyys (approved):
LGTM

math/fixed_point/sources/internal/common.move
ericnordelo: Let's consider doing that later but avoid delaying the PR more.

immrsd (approved):
LGTM!

PR #282, opened 2026-03-30
test: change expect helper function into macro

0xNeshi:
`assert_eq!` macro can be used directly on fixed point types, so there's no need for `expect` wrapper.

https://move-book.com/reference/equality


ericnordelo (dismissed):
LGTM!

bidzyyys (changes requested):
Why can't we have `assert!(left.eq(right))`?

bidzyyys (approved):
Discussed on Slack.

bidzyyys (changes requested):
Use `assert_eq` as discussed on Slack.

ericnordelo (approved):
LGTM!

bidzyyys (approved):
LGTM

Issue #136, opened 2026-02-05
[Design]: Rate Limiter module


bidzyyys:
### What is the feature you would like to see?

Implement Rate Limiter module, similar to [this implementation](https://github.com/MystenLabs/deepbookv3/blob/main/packages/deepbook_margin/sources/rate_limiter.move), but:
- make it more generic (in a "library way")
- add a possibility to parametrise it 
- think about ownership & access control during the design

Implementation steps:
- [x] Create a design
- [x] Discuss with the internal team
- [x] Discuss with Sui DeFi Team

PR #284, opened 2026-03-31
Rate limiter proposal Examples

0xNeshi:
Towards #136 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog


rate_limiter_proposal/token_bucket_example/sources/vault.move
```diff
@@ -0,0 +1,97 @@
...
+#[error(code = 1)]
+const EWrongPolicy: vector<u8> = "Wrong policy";
+
+public struct Vault has key, store {
+    id: UID,
+    policy_id: ID,
+    balance: Balance<SUI>,
+}
+
+public struct WithdrawTag has copy, drop, store {}
```
ericnordelo, line 20: This shouldn't need store, and I'm not sure about other abilities.

rate_limiter_proposal/token_bucket_example/sources/vault.move
```diff
@@ -0,0 +1,97 @@
...
+        initial_tokens,
+        ctx,
+    );
+    let vault = Vault {
+        id: object::new(ctx),
+        policy_id: object::id(&policy),
+        balance: coin::into_balance(initial_coin),
+    };
+    transfer::share_object(vault);
+    transfer::public_share_object(policy);
```
ericnordelo, line 45: We should be able to do share policies automatically on creation so users won't forget, maybe we introduce a hot potatto pattern for configuration if we need one,

rate_limiter_proposal/token_bucket_example/sources/vault.move
```diff
@@ -0,0 +1,97 @@
...
+    self: &mut Vault,
+    policy: &token_bucket::Policy<WithdrawTag>,
+    deposit_coin: Coin<SUI>,
+    clock: &Clock,
+    ctx: &mut TxContext,
+): token_bucket::State<WithdrawTag> {
+    assert_policy(self, policy);
+    balance::join(&mut self.balance, coin::into_balance(deposit_coin));
+    let withdrawer = tx_context::sender(ctx);
+    token_bucket::create_for_address<WithdrawTag>(
```
ericnordelo, line 58: This is probably not the best example for a vault, since it would be rate limiting per user, and the intention of rate limiter is to limit aggregate withdraws, which this won't do if there is massive withdraw from separate accounts.

0xNeshi: this would still be supported, the difference would be that the `State` (rate limiting object) would be created only once, instead of per deposit

0xNeshi: I can update the example, or provide an alternative implementation

ericnordelo: Yes, but my comment was incomplete, sorry. I sent it to slack but forgot to mentioning in here: How does the vault update the policy on this scenario if every separated state has the policy id and are owned by users?

0xNeshi: will address that in the examples repo

0xNeshi:
Closing as the examples are moved to separate repo:
https://github.com/0xNeshi/rate-limiter-proposal

PR #288, opened 2026-04-03
feat: Support both checked and unchecked versions of left and right shift

### LINKED ISSUE #270: [L-02]: Unsigned Fixed-Point Left Shift Silently Truncates on Overflow

0xNeshi:
Resolves #270 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog


math/fixed_point/sources/ud30x9/ud30x9_base.move
```diff
@@ -203,9 +225,9 @@ public fun is_zero(x: UD30x9): bool {
 /// #### Returns
 /// - Zero if `bits >= 128` (all bits shifted out).
 /// - Otherwise, the result of shifting the `x`'s raw bits left by `bits`.
-public fun lshift(x: UD30x9, bits: u8): UD30x9 {
+public fun unchecked_lshift(x: UD30x9, bits: u8): UD30x9 {
```
ericnordelo, line 237: What do you think about using `wlshift`? Three reasons, the first one is that is shorter avoiding complex long lines, second is that lshift already shortens left and wl would be (with right documentation) wrapped_left, third that we use checked semantics when we return options instead of aborting, and unchecked in this context goes against that semantic.

0xNeshi: Tbh I'd go as far as following Rust's naming convention, which is already a widely understood "standard" - `shl / shr` and `unchecked_shl / unchecked_shr`, especially since our implementation's behavior matches Rust's.

To get back to your rename suggestion, I don't think `wlshift` is a good idea. The `unchecked_lshift` name is immediately self-documenting and conventional in the ecosystem (as well as in our own library, which uses the same prefix throughout), while `wlshift` is neither, and would force the reader to read docs to know what it does.

Line length (16 chars) shouldn't be a concern, and if it is, devs can always alias. IMO this isn't a strong reason to sacrifice readability.

> ...third that we use checked semantics when we return options instead of aborting, and unchecked in this context goes against that semantic.

Yes, `checked_` is used when options are returned, just the name (e.g. `lshift`) is usually used with aborting/panicking, but `unchecked_` usually does not entail options, so I'm not sure how using this prefix goes against the semantic. Especially, as I mentioned above, our implementation's behavior matches Rust's.

0xNeshi: Side note, these operations can only _technically_ be considered wrapping (due to the result being modulo), but in reality they are truncating.

bidzyyys (approved):

CHANGELOG.md
```diff
@@ -12,10 +12,12 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 #### Added
 
-- `SD29x9::rem` function for truncated remainder semantics (sign follows the dividend) (#301)
+- Added `UD30x9::unchecked_lshift` and `UD30x9::unchecked_rshift` with truncating / zero-return behavior for shift operations. (#288)
 
 #### Changed (Breaking)
 
+- `UD30x9::lshift` now aborts on overflow, and both `lshift` and `rshift` now abort on invalid shift sizes. (#288)
+- `SD29x9::rem` function for truncated remainder semantics (sign follows the dividend) (#301)
```
immrsd: I believe the changelog entry for `SD29x9::rem` should be kept in the "Added" section

0xNeshi:
```suggestion
- `SD29x9::rem` function for truncated remainder semantics (sign follows the dividend) (#301)
- Added `UD30x9::unchecked_lshift` and `UD30x9::unchecked_rshift` with truncating / zero-return behavior for shift operations. (#288)

#### Changed (Breaking)

- `UD30x9::lshift` now aborts on overflow, and both `lshift` and `rshift` now abort on invalid shift sizes. (#288)
```

CHANGELOG.md
```diff
@@ -21,6 +23,7 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 - Removed bitwise operations from `SD29x9`.
 - Removed public module `casting_u128`; use `ud30x9::wrap` and `sd29x9::wrap` directly for raw casts. (#264)
+- `UD30x9::lshift` now aborts on overflow, and both `lshift` and `rshift` now abort on invalid shift sizes. (#288)
 - `SD29x9::rem` function for truncated remainder semantics (sign follows the dividend) (#301)
```
0xNeshi:
```suggestion
```

math/fixed_point/sources/ud30x9/ud30x9_base.move
ericnordelo: Great points, let's keep them.

immrsd (approved):
LGTM

ericnordelo (approved):

PR #300, opened 2026-04-06
fix: expose public `decompose` (L-05)

### LINKED ISSUE #273: [L-05]: Decompose Is Not Publicly Exposed, Forcing Users to Validate Independently

immrsd:
Summary
                                                                                                                         
- Expose `SD29x9::decompose` as a public function returning `(bool, u128)` so external callers can extract sign and
magnitude through the canonical decomposition path instead of reimplementing the logic
- Rename the internal helper to `decompose_bits` to avoid name collision

### DIFF
--- a/math/fixed_point/sources/sd29x9/sd29x9_base.move
+++ b/math/fixed_point/sources/sd29x9/sd29x9_base.move
@@ -36,7 +36,7 @@ const ECannotBeConvertedToUD30x9: vector<u8> = "Value cannot be converted to UD3
 /// #### Aborts
 /// - Aborts if `x` is negative.
 public fun into_UD30x9(x: SD29x9): UD30x9 {
-    let Components { neg, mag } = decompose(x.unwrap());
+    let Components { neg, mag } = decompose_bits(x.unwrap());
     assert!(!neg, ECannotBeConvertedToUD30x9);
     ud30x9::wrap(mag as u128)
 }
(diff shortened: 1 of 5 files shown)


math/fixed_point/sources/sd29x9/sd29x9_base.move
```diff
@@ -140,6 +140,20 @@ public fun ceil(x: SD29x9): SD29x9 {
...
+///
+/// #### Returns
+/// - A tuple `(is_negative, magnitude)` where:
+///   - `is_negative`: `true` if `x` is negative.
+///   - `magnitude`: The absolute value of `x` as a `u128`.
+public fun decompose(x: SD29x9): (bool, u128) {
+    let Components { neg, mag } = decompose_bits(x.unwrap());
+    (neg, mag as u128)
+}
+
```
ericnordelo, line 156: I'd keep decompose private. It's an internal encoding helper, not really a user-facing operation, and the min value makes it especially awkward as public API because it exposes a representation detail that does not round-trip cleanly through the normal constructors. The main issue is the min value edge case, decomposing min gives you a magnitude that cannot be reconstructed through the normal public constructor path, so exposing decompose would leak an internal invariant without giving users a clean or symmetric way to work with it.

If we need this publicly, I'd rather expose a higher-level helper designed around the use case instead of leaking the raw internals, and I would leave it for a future iteration, as it needs careful design about the API, is not a quick fix.

0xNeshi (approved):
Tentative approval, see https://github.com/OpenZeppelin/contracts-sui/pull/300/changes#r3040415220

immrsd:
Decided to postpone adding the `decompose` function and consider this function in the next milestone.

PR #313, opened 2026-05-04
Add RBAC

### LINKED ISSUE #12: [Feature]: Implement Access Control
Implement role-based access control contract based on [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControl.sol) and prior research in #196.

### LINKED ISSUE #196: [Design]: Access Control
- [ ] Include design and features about [Access Manager](https://docs.openzeppelin.com/contracts/5.x/access-control#access-management) (#192).

ericnordelo:
Resolves #12 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog

## Summary by CodeRabbit

* **New Features**
  * Role-based access control with managed role membership, per-role admin wiring, and role-issued membership credentials.
  * Timelocked root admin flows: schedule/accept/cancel transfers and renounces; separate timelocked flow to schedule/accept/cancel admin delay changes; visibility into pending actions.

* **Documentation**
  * Expanded access docs with an overview, usage flow, and module snapshot for the new access control features.

### DIFF
--- a/contracts/access/README.md
+++ b/contracts/access/README.md
@@ -1,8 +1,8 @@
 # Access Package
 
-Transfer policies for safely handing objects to new owners.
+Authorization primitives for Sui Move libraries, standards, and protocols: role-based access control plus transfer policies for safely handing objects to new owners.
 
-Both modules accept any `T: key + store` object, but the primary use case is wrapping **capabilities** — admin caps, treasury caps, and similar authority tokens — where a misdirected transfer could be irreversible.
+Transfer-policy modules accept any `T: key + store` object, but the primary use case is wrapping **capabilities** — admin caps, treasury caps, and similar authority tokens — where a misdirected transfer could be irreversible. The access-control module gates protocol functions by typed role witnesses without per-call ID pinning.
 
 ---
 
@@ -10,8 +10,30 @@ Both modules accept any `T: key + store` object, but the primary use case is wra
 
 | Module | Summary |
 |--------|---------|
+| `access_control` | Role-based access control with self-validating `Auth<R>` capabilities, OTW-pinned singleton registry, and timelocked root role transfer. |
 | `two_step_transfer` | Wraps an object behind an initiate_transfer/accept_transfer flow to avoid accidental handoffs. |
 | `delayed_transfer` | Enforces a minimum delay between scheduling and executing transfers or unwrapping. |
+
+---
+
+## `access_control` at a Glance
+
+The library guarantees two structural invariants: (1) **one registry per module per publish** (enforced by requiring a One-Time Witness in `new<RootRole>`); (2) **only home-module roles** (every write path checks that the role type's defining module matches the root role's, using `with_original_ids` so role types added in package upgrades are accepted). Together these make `Auth<R>` a self-validating capability — action functions take `&Auth<R>` directly with no body checks.
+
+1. **Deploy**
+   - In your module's `init`, call `access_control::new<MY_OTW>(otw, default_admin_delay_ms, ctx)` to mint the singleton `AccessControl<MY_OTW>`. The publishing transaction's sender automatically becomes the root role holder. Share or embed the registry as your protocol requires.
+2. **Wire roles**
+   - Define phantom marker types in the same module as your OTW (e.g. `public struct AdminRole {}`, `public struct OperatorRole {}`).
+   - Build the admin hierarchy with `set_role_admin<RootRole, Role, AdminRole>` — every fresh role's admin defaults to the root role; rewire as needed.
+   - Use `grant_role<RootRole, R>`, `revoke_role<RootRole, R>`, and `renounce_role<RootRole, R>` to manage membership. Granting / revoking is gated by holding `R`'s admin role; renounce is self-only.
+3. **Issue capabilities**
+   - `new_auth<RootRole, R>(&registry, ctx)` mints `Auth<R>` for `ctx.sender()` if they currently hold `R`.
+   - Pass `&Auth<R>` to gated business actions in the same PTB. Because the registry is singleton-per-publish and only home-module roles can ever be registered in it, any `Auth<R>` in scope is unforgeable in context — no body checks needed.
+4. **Rotate or relinquish root**
+   - `begin_default_admin_transfer(&mut registry, new_admin, &clock, ctx)` schedules a transfer from the current root holder.
+   - After `default_admin_delay_ms` elapses, the pending admin calls `accept_default_admin_transfer(&mut registry, &clock, ctx)`. The rotation is atomic: the old admin is revoked and the new admin is granted in one transaction.
+   - The current root holder can call `cancel_default_admin_transfer(&mut registry, ctx)` to back out at any time before acceptance.
+   - Renouncing the root role when you are the last holder makes the registry permanently unmanageable — used deliberately as a final lock-in step.
 
 ---
 
--- a/contracts/access/README.md
+++ b/contracts/access/README.md
@@ -1,4 +1,4 @@
-# OpenZeppelin Access for Sui
+# `openzeppelin_access`
 
 Role-based authorization and controlled ownership-transfer primitives for Sui Move protocols.
 
@@ -24,6 +24,12 @@ openzeppelin_access = { r.mvr = "@openzeppelin-move/access" }
 Use `access_control` when your package needs operational roles such as admins, treasurers, guardians, pausers, keepers, or governance executors.
 
 The registry is rooted in your module's One-Time Witness (OTW). Roles must be defined in the same module as that OTW, and `Auth<Role>` values can be minted in PTBs as typed authorization proofs.
+
+Initialize the registry from the module's `init` function when the package is
+first published. Sui does not run `init` for modules added in a later package
+upgrade, so an already-published package cannot adopt AccessControl by adding a
+new module that expects its OTW to be delivered. Publish a fresh package with
+its own initializing module and AccessControl registry instead.
 
 ```move
 use openzeppelin_access::access_control::{Self, AccessControl, Auth};
--- a/contracts/access/README.md
+++ b/contracts/access/README.md
@@ -1,67 +1,83 @@
-# Access Package
+# OpenZeppelin Access for Sui
 
-Authorization primitives for Sui Move libraries, standards, and protocols: role-based access control plus transfer policies for safely handing objects to new owners.
+Role-based authorization and controlled ownership-transfer primitives for Sui Move protocols.
 
-Transfer-policy modules accept any `T: key + store` object, but the primary use case is wrapping **capabilities** — admin caps, treasury caps, and similar authority tokens — where a misdirected transfer could be irreversible. The access-control module gates protocol functions by typed role witnesses without per-call ID pinning.
+The `openzeppelin_access` package helps protocols protect privileged functions, shared state, admin capabilities, treasury capabilities, and governance-controlled operations.
 
----
+## Install
 
-## Module Snapshot
+```toml
+[dependencies]
+openzeppelin_access = { r.mvr = "@openzeppelin-move/access" }
+```
 
-| Module | Summary |
-|--------|---------|
-| `access_control` | Role-based access control with self-validating `Auth<R>` capabilities, OTW-pinned singleton registry, and timelocked root role transfer. |
-| `two_step_transfer` | Wraps an object behind an initiate_transfer/accept_transfer flow to avoid accidental handoffs. |
-| `delayed_transfer` | Enforces a minimum delay between scheduling and executing transfers or unwrapping. |
+## Modules
 
----
+| Module | Use it when |
+| --- | --- |
+| [`access_control`](https://docs.openzeppelin.com/contracts-sui/1.x/access-control) | Authority is spread across multiple roles or actors, especially for shared protocol state and privileged functions. |
+| [`two_step_transfer`](https://docs.openzeppelin.com/contracts-sui/1.x/two-step-transfer) | A single-owned privileged object can transfer immediately, but the recipient should explicitly accept first. |
+| [`delayed_transfer`](https://docs.openzeppelin.com/contracts-sui/1.x/delayed-transfer) | A single-owned privileged object should not transfer or unwrap until a visible delay has elapsed. |
 
-## `access_control` at a Glance
+## Role-Based Access Control
 
-The library guarantees two structural invariants: (1) **one registry per module per publish** (enforced by requiring a One-Time Witness in `new<RootRole>`); (2) **only home-module roles** (every write path checks that the role type's defining module matches the root role's, using `with_original_ids` so role types added in package upgrades are accepted). Together these make `Auth<R>` a self-validating capability — action functions take `&Auth<R>` directly with no body checks.
+Use `access_control` when your package needs operational roles such as admins, treasurers, guardians, pausers, keepers, or governance executors.
 
-1. **Deploy**
-   - In your module's `init`, call `access_control::new<MY_OTW>(otw, default_admin_delay_ms, ctx)` to mint the singleton `AccessControl<MY_OTW>`. The publishing transaction's sender automatically becomes the root role holder. Share or embed the registry as your protocol requires.
-2. **Wire roles**
-   - Define phantom marker types in the same module as your OTW (e.g. `public struct AdminRole {}`, `public struct OperatorRole {}`).
-   - Build the admin hierarchy with `set_role_admin<RootRole, Role, AdminRole>` — every fresh role's admin defaults to the root role; rewire as needed.
-   - Use `grant_role<RootRole, R>`, `revoke_role<RootRole, R>`, and `renounce_role<RootRole, R>` to manage membership. Granting / revoking is gated by holding `R`'s admin role; renounce is self-only.
-3. **Issue capabilities**
-   - `new_auth<RootRole, R>(&registry, ctx)` mints `Auth<R>` for `ctx.sender()` if they currently hold `R`.
-   - Pass `&Auth<R>` to gated business actions in the same PTB. Because the registry is singleton-per-publish and only home-module roles can ever be registered in it, any `Auth<R>` in scope is unforgeable in context — no body checks needed.
-4. **Rotate or relinquish root**
-   - `begin_default_admin_transfer(&mut registry, new_admin, &clock, ctx)` schedules a transfer from the current root holder.
-   - After `default_admin_delay_ms` elapses, the pending admin calls `accept_default_admin_transfer(&mut registry, &clock, ctx)`. The rotation is atomic: the old admin is revoked and the new admin is granted in one transaction.
-   - The current root holder can call `cancel_default_admin_transfer(&mut registry, ctx)` to back out at any time before acceptance.
-   - Renouncing the root role when you are the last holder makes the registry permanently unmanageable — used deliberately as a final lock-in step.
+The registry is rooted in your module's One-Time Witness (OTW). Roles must be defined in the same module as that OTW, and `Auth<Role>` values can be minted in PTBs as typed authorization proofs.
 
----
+```move
+use openzeppelin_access::access_control::{Self, AccessControl, Auth};
+```
 
-## `two_step_transfer` at a Glance
+Start with the [Role Based Access Control guide](https://docs.openzeppelin.com/contracts-sui/1.x/guides/access-control) for a full implementation walkthrough with publishing, upgrades, and PTBs.
 
-1. **Wrap**
-   - Call `wrap<T>` to place an object under a `TwoStepTransferWrapper<T>`. The wrapper owns the object via a dynamic object field.
-2. **Borrow**
-   - Use `borrow`, `borrow_mut`, or `borrow_val`/`return_val` to read or temporarily mutate the wrapped object without changing ownership.
-   - While a transfer is pending, the current owner can use `request_borrow_val`/`request_return_val` to temporarily access the wrapper and its inner object through the shared request.
-3. **Transfer**
-   - The current owner calls `initiate_transfer` to emit `TransferInitiated`, create a shared `PendingOwnershipTransfer`, and TTO the wrapper to it.
-   - The prospective owner calls `accept_transfer` with the request + receiving ticket to accept, or the current owner calls `cancel_transfer` to reclaim the wrapper.
-4. **Unwrap**
-   - Owners can reclaim the underlying object immediately via `unwrap`, destroying the wrapper.
+Relevant docs:
 
----
+- [RBAC package page](https://docs.openzeppelin.com/contracts-sui/1.x/access-control)
+- [Implement a module with RBAC](https://docs.openzeppelin.com/contracts-sui/1.x/guides/access-control#implementing-a-new-module)
+- [Upgrade an existing package](https://docs.openzeppelin.com/contracts-sui/1.x/guides/access-control#upgrading-an-existing-package)
+- [Root role operations](https://docs.openzeppelin.com/contracts-sui/1.x/guides/access-control#root-role-operations)
+- [`access_control` API reference](https://docs.openzeppelin.com/contracts-sui/1.x/api/access#access_control)
 
-## `delayed_transfer` at a Glance
+## Controlled Object Transfers
 
-1. **Wrap**
-   - `wrap<T>(obj, min_delay_ms, recipient, ctx)` creates a `DelayedTransferWrapper<T>`, stores the object under it, and transfers the wrapper to `recipient`.
-2. **Borrow**
-   - Access the wrapped object through `borrow`, `borrow_mut`, or the `borrow_val` / `return_val` pair for temporary moves.
-3. **Schedule**
-   - Call `schedule_transfer` with a recipient and clock to emit `TransferScheduled`. The wrapper records the current owner from `ctx.sender()`.
-   - Call `schedule_unwrap` to plan a delayed unwrap; the wrapper emits `UnwrapScheduled`.
-4. **Execute or Cancel**
-   - After `clock.timestamp_ms() >= execute_after_ms`, call `execute_transfer` to emit `OwnershipTransferred` and move the wrapper to the scheduled recipient.
-   - For scheduled unwraps, call `unwrap` to emit `UnwrapExecuted`, delete the wrapper, and recover the underlying object.
-   - Use `cancel_schedule` to cancel a pending transfer or unwrap before execution.
+Use the transfer modules for single-owned privileged objects, especially capability objects.
+
+### Two-step transfer
+
+`two_step_transfer` wraps an object and requires the intended recipient to accept before ownership moves.
+
+```move
+use openzeppelin_access::two_step_transfer;
+```
+
+Docs:
+
+- [Two-Step Transfer guide](https://docs.openzeppelin.com/contracts-sui/1.x/two-step-transfer)
+- [`two_step_transfer` API reference](https://docs.openzeppelin.com/contracts-sui/1.x/api/access#two_step_transfer)
+
+### Delayed transfer
+
+`delayed_transfer` wraps an object and enforces a minimum delay before transfer or unwrap execution.
+
+```move
+use openzeppelin_access::delayed_transfer;
+```
+
+Docs:
+
+- [Delayed Transfer guide](https://docs.openzeppelin.com/contracts-sui/1.x/delayed-transfer)
+- [`delayed_transfer` API reference](https://docs.openzeppelin.com/contracts-sui/1.x/api/access#delayed_transfer)
+
+## Security Notes
+
+- `access_control` root-role changes use delayed flows. Do not try to grant, revoke, or renounce the root role with ordinary role-management calls.
+- Role grants and root transfers reject `@0x0`. Use the delayed root-renounce flow when the goal is to intentionally lock a registry.
+- `two_step_transfer` records `ctx.sender()` as cancel authority. Avoid using it directly in shared-object executor flows unless signer identity is intentionally the cancel authority.
+- `delayed_transfer` recipients should be wallet addresses, not object IDs, unless your protocol explicitly supports recovery from transfer-to-object custody.
+
+## Learn More
+
+- [Access package overview](https://docs.openzeppelin.com/contracts-sui/1.x/access)
+- [Access API reference](https://docs.openzeppelin.com/contracts-sui/1.x/api/access)
+- [OpenZeppelin Contracts for Sui](https://docs.openzeppelin.com/contracts-sui)
--- a/contracts/access/sources/access_control.move
+++ b/contracts/access/sources/access_control.move
@@ -1,0 +1,572 @@
+/// Role-based access control for Sui Move protocols.
+///
+/// ### Structural invariants
+///
+/// 1. **One AccessControl per module per publish.** `new<RootRole>` only accepts
+///    a One-Time Witness as `RootRole`. The Sui VM produces exactly one OTW
+///    value of a given type per package publish (passed into the consumer
+///    module's `init`), so a module can stand up at most one registry of its
+///    own type, ever.
+///
+/// 2. **Only home-module roles.** Every write path (`grant_role`, `revoke_role`,
+///    `renounce_role`, `set_role_admin`, `new_auth`) checks that the role
+///    type's home module matches the root role's home module. Foreign role
+///    types are rejected at the boundary; they cannot be introduced into the
+///    bag. The check uses `type_name::with_original_ids`, which compares the
+///    package's *original* publish address — so role types introduced in
+///    later upgrades of the same package pass the check too. The home-module
+///    rule restricts roles to the same package, not to the same package
+///    version.
+///
+/// Together these guarantees make `Auth<R>` a self-validating capability.
+/// `R` can only live in the registry of its home module, that registry is
+/// unique per publish, and the only path to mint `Auth<R>` is `new_auth`
+/// against that registry. Action functions can take `&Auth<R>` directly with
+/// no body checks.
+///
+/// ### Idiomatic usage
+///
+/// ```move
+/// module my_protocol::my_protocol;
+///
+/// public struct MY_PROTOCOL has drop {}    // OTW; serves as the root role
+/// public struct AdminRole {}               // additional role, same module
+/// public struct OperatorRole {}            // additional role, same module
+///
+/// fun init(otw: MY_PROTOCOL, ctx: &mut TxContext) {
+///     let mut registry = access_control::new(otw, 86_400_000, ctx);
+///     access_control::grant_role<_, AdminRole>(
+///         &mut registry, ctx.sender(), ctx,
+///     );
+///     access_control::set_role_admin<_, OperatorRole, AdminRole>(
+///         &mut registry, ctx,
+///     );
+///     transfer::public_share_object(registry);
+/// }
+/// ```
+module openzeppelin_access::access_control;
+
+use std::type_name::{TypeName, with_original_ids};
+use sui::bag::{Self, Bag};
+use sui::clock::Clock;
+use sui::event;
+use sui::types;
+use sui::vec_set::{Self, VecSet};
+
+// === Errors ===
+
+/// Caller does not hold the required role.
+#[error(code = 0)]
+const EUnauthorized: vector<u8> = "Caller does not have the required role";
+
+/// A write operation was attempted on the protected root role.
+#[error(code = 1)]
+const ECannotManageRootRole: vector<u8> =
+    "Use begin_default_admin_transfer to change the root role holder";
+
+/// `accept_default_admin_transfer` or `cancel_default_admin_transfer` called
+/// with no pending transfer.
+#[error(code = 2)]
+const ENoPendingAdminTransfer: vector<u8> = "No pending default admin transfer";
+
+/// `accept_default_admin_transfer` called by an address other than the pending
+/// new admin.
+#[error(code = 3)]
+const ENotPendingAdmin: vector<u8> = "Caller is not the pending admin";
+
+/// `accept_default_admin_transfer` called before the configured delay has
+/// elapsed.
+#[error(code = 4)]
+const EDelayNotElapsed: vector<u8> = "Admin transfer delay has not elapsed";
+
+/// `renounce_role` called with `account != ctx.sender()`.
+#[error(code = 5)]
+const ECannotRenounceForOtherAccount: vector<u8> = "Can only renounce role for own account";
+
+/// `new` called with `default_admin_delay_ms` above `MAX_DEFAULT_ADMIN_DELAY_MS`.
+#[error(code = 6)]
+const EDelayTooLarge: vector<u8> = "Admin delay exceeds the maximum allowed value";
+
+/// `new` called with a value that is not a genuine One-Time Witness.
+#[error(code = 7)]
+const ENotOneTimeWitness: vector<u8> = "Root role must be a One-Time Witness";
+
+/// A role type from a module other than the root role's home module was used
+/// in a write path.
+#[error(code = 8)]
+const EForeignRole: vector<u8> = "Role type must be defined in the same module as the root role";
+
+// === Constants ===
+
+/// Maximum value for `default_admin_delay_ms` passed to `new()`.
+/// Value: 30 days in milliseconds.
+const MAX_DEFAULT_ADMIN_DELAY_MS: u64 = 30 * 24 * 60 * 60 * 1_000;
+
+// === Structs ===
+
+/// Self-validating typed proof of role membership.
+///
+/// `Auth<R>` is minted exclusively by `new_auth` against the unique
+/// `AccessControl<RootRole>` whose home module matches `R`'s. Because that
+/// registry is singleton-per-publish (Invariant 1) and only home-module roles
+/// can ever be registered in it (Invariant 2), every `Auth<R>` that exists was
+/// produced by that one registry, against a sender that genuinely held `R`.
+/// Consumers can take `&Auth<R>` as authorization without any body checks.
+public struct Auth<phantom R> has drop {
+    addr: address,
+}
+
+/// Central access control registry.
+///
+/// `RootRole` is the consuming module's One-Time Witness. The phantom parameter
+/// pins the registry's identity at the type level so the same-module check in
+/// every write path can reject foreign role types statically attached to this
+/// registry.
+public struct AccessControl<phantom RootRole> has key, store {
+    id: UID,
+    /// Per-role membership and admin mapping. Keyed by `TypeName`.
+    /// Entries are created lazily on first grant or `set_role_admin`.
+    roles: Bag,
+    /// `TypeName` of the root role (= `RootRole`'s OTW). Direct grant/revoke
+    /// on this role is blocked; use the timelocked transfer flow.
+    protected_root: TypeName,
+    /// Pending root role transfer.
+    pending_default_admin: Option<PendingAdminTransfer>,
+    /// Minimum delay (ms) for root role transfer. Set at creation; immutable.
+    default_admin_delay_ms: u64,
+}
+
+/// Per-role state: current members and the role that administers it.
+public struct RoleData has store {
+    members: VecSet<address>,
+    /// Defaults to the root role; changeable via `set_role_admin`.
+    admin_role: TypeName,
+}
+
+/// Snapshot of a pending root role transfer.
+public struct PendingAdminTransfer has drop, store {
+    new_admin: address,
+    execute_after_ms: u64,
+}
+
+// === Events ===
+//
+// The `role` field on every event is a `TypeName` — it embeds the defining
+// package address and module name of the role struct, so events from
+// independently-published consumer protocols are distinguishable off-chain.
+
+public struct RoleGranted has copy, drop {
+    role: TypeName,
+    account: address,
+    sender: address,
+}
+
+public struct RoleRevoked has copy, drop {
+    role: TypeName,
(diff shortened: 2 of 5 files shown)


contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+/// enforces Invariant 1 (one registry per module per publish). The
+/// transaction sender automatically becomes the root role holder.
+///
+/// The expected call site is the consumer module's `init` function, where
+/// the VM has already produced exactly one OTW value.
+///
+/// #### Aborts
+/// - `ENotOneTimeWitness` if `otw` is not a true One-Time Witness.
+/// - `EDelayTooLarge` if `default_admin_delay_ms` exceeds `MAX_DEFAULT_ADMIN_DELAY_MS`.
+public fun new<RootRole: drop>(
```
0xNeshi, line 302: Missing `#### Parameters` doc section in most/all function docs

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+///
+/// public struct MY_PROTOCOL has drop {}    // OTW; serves as the root role
+/// public struct AdminRole {}               // additional role, same module
+/// public struct OperatorRole {}            // additional role, same module
+///
+/// fun init(otw: MY_PROTOCOL, ctx: &mut TxContext) {
+///     let mut registry = access_control::new(otw, 86_400_000, ctx);
+///     access_control::grant_role<_, AdminRole>(
+///         &mut registry, ctx.sender(), ctx,
+///     );
```
0xNeshi: Use method call syntax (`registry.grant_role`), instead of fully qualified syntax everywhere

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+
+/// Self-validating typed proof of role membership.
+///
+/// `Auth<R>` is minted exclusively by `new_auth` against the unique
+/// `AccessControl<RootRole>` whose home module matches `R`'s. Because that
+/// registry is singleton-per-publish (Invariant 1) and only home-module roles
+/// can ever be registered in it (Invariant 2), every `Auth<R>` that exists was
+/// produced by that one registry, against a sender that genuinely held `R`.
+/// Consumers can take `&Auth<R>` as authorization without any body checks.
+public struct Auth<phantom R> has drop {
```
```suggestion
public struct Auth<phantom Role> has drop {
```
Consider using a full name for the generic param for increased clarity, and alignment with [AccessControl<phantom RootRole>](https://github.com/OpenZeppelin/contracts-sui/pull/313/changes#diff-a8d2bcee6cb0a9f5c79d7dabcedc47433f84979e57d3a518a9c07108f32947c5R125)

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+                admin_role: root_type,
+            },
+        );
+
+    event::emit(RoleGranted { role: root_type, account: sender, sender });
+
+    ac
+}
+
+// === Internal Helpers ===
```
0xNeshi: Internal helpers should go after all public functions. Run [move-quality](https://github.com/OpenZeppelin/sui-claude-skills/blob/main/move-quality/SKILL.md#section-ordering) skill to fix issues.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+            && root.module_string() == role.module_string(),
+        EForeignRole,
+    );
+}
+
+// === Role Queries ===
+
+/// Returns `true` if `account` is a member of role `R`.
+public fun has_role<RootRole, R>(ac: &AccessControl<RootRole>, account: address): bool {
+    has_role_by_name(ac, with_original_ids<R>(), account)
```
```suggestion
    ac.has_role_by_name(with_original_ids<R>(), account)
```
Use method call syntax everywhere

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+    execute_after_ms: u64,
+}
+
+// === Events ===
+//
+// The `role` field on every event is a `TypeName` — it embeds the defining
+// package address and module name of the role struct, so events from
+// independently-published consumer protocols are distinguishable off-chain.
+
+public struct RoleGranted has copy, drop {
```
0xNeshi, line 218: Missing docs on all events.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+public struct PendingAdminTransfer has drop, store {
+    new_admin: address,
+    execute_after_ms: u64,
+}
+
+// === Events ===
+//
+// The `role` field on every event is a `TypeName` — it embeds the defining
+// package address and module name of the role struct, so events from
+// independently-published consumer protocols are distinguishable off-chain.
```
0xNeshi: This would be more useful info if it appeared when hovering on every relevant event, i.e. if it was copy/pasted in all events' docs.

contracts/access/sources/access_control.move
0xNeshi, line 302: Missing `#### Returns` section in most/all relevant functions

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+    let role_data = ac.roles.borrow_mut<_, RoleData>(role_name);
+    if (!role_data.members.contains(&account)) return;
+
+    role_data.members.remove(&account);
+    event::emit(RoleRevoked { role: role_name, account, sender: ctx.sender() });
+}
+
+/// Set the admin role of `Role` to `AdminRole`. Both must be home-module
+/// roles. Caller must hold the current admin role of `Role`. Blocked on the
+/// root role as the subject.
```
0xNeshi: > Blocked on the root role as the subject.

Unclear what is meant here, but probably that `Role = RootRole` will abort, which is already described in `aborts` section. 
Either reword or remove completely.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+
+    if (!ac.roles.contains(role_name)) {
+        ac
+            .roles
+            .add(
+                role_name,
+                RoleData { members: vec_set::empty(), admin_role: new_admin_name },
+            );
+    } else {
+        ac.roles.borrow_mut<_, RoleData>(role_name).admin_role = new_admin_name;
```
```suggestion
    if (ac.roles.contains(role_name)) {
        ac.roles.borrow_mut<_, RoleData>(role_name).admin_role = new_admin_name;
    } else {
        ac
            .roles
            .add(
                role_name,
                RoleData { members: vec_set::empty(), admin_role: new_admin_name },
            );
```
nit: invert branches to avoid negation in `if` (possible tiny performance improvement)

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+/// Caller must hold the root role. The transfer cannot be accepted until
+/// `default_admin_delay_ms` has elapsed. An existing pending transfer is
+/// overwritten — the caller can correct a wrong recipient without cancelling.
+public fun begin_default_admin_transfer<RootRole>(
+    ac: &mut AccessControl<RootRole>,
+    new_admin: address,
+    clock: &Clock,
+    ctx: &mut TxContext,
+) {
+    assert!(has_role_by_name(ac, ac.protected_root, ctx.sender()), EUnauthorized);
```
0xNeshi: Missing zero-address check on `new_admin`

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+/// Aborts with `EUnauthorized` if `account` does not hold role `R`.
+public fun assert_role<RootRole, R>(ac: &AccessControl<RootRole>, account: address) {
+    assert!(has_role<RootRole, R>(ac, account), EUnauthorized);
+}
+
+// === Role Management ===
+
+/// Grant role `R` to `account`. Caller must hold the admin role of `R`.
+/// `R` must be defined in the same module as `RootRole` (Invariant 2).
+/// Blocked on the root role. No-op if `account` already holds `R`.
```
```suggestion
/// No-op if `account` already holds `R`.
```

As before, slightly unclear. Can be removed, as `Aborts` already documents this case.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+
+    let role_data = ac.roles.borrow_mut<_, RoleData>(role_name);
+    if (role_data.members.contains(&account)) return;
+
+    role_data.members.insert(account);
+    event::emit(RoleGranted { role: role_name, account, sender: ctx.sender() });
+}
+
+/// Revoke role `R` from `account`. Caller must hold the admin role of `R`.
+/// Blocked on the root role. No-op if `account` does not hold `R`.
```
```suggestion
/// No-op if `account` does not hold `R`.
```

As before, slightly unclear. Can be removed, as `Aborts` already documents this case.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+    assert!(account == ctx.sender(), ECannotRenounceForOtherAccount);
+    assert_home_module<RootRole, R>();
+
+    let role_name = with_original_ids<R>();
+    if (!ac.roles.contains(role_name)) return;
+
+    let role_data = ac.roles.borrow_mut<_, RoleData>(role_name);
+    if (!role_data.members.contains(&account)) return;
+
+    role_data.members.remove(&account);
```
0xNeshi, line 492: If `role_data.members.is_empty()`, should we remove the role from `ac.roles`?
One consequence of removing zero-member roles is that `get_role_admin` would go back to returning root role.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+    let root_type = ac.protected_root;
+
+    // Root role has at most one holder; find and revoke atomically.
+    let maybe_old_admin = {
+        let keys = ac.roles.borrow<_, RoleData>(root_type).members.keys();
+        if (keys.is_empty()) option::none() else option::some(keys[0])
+    };
+
+    if (maybe_old_admin.is_some()) {
+        let old_admin = *maybe_old_admin.borrow();
```
```suggestion
    let keys = ac.roles.borrow<_, RoleData>(root_type).members.keys();
    if (!keys.is_empty()) {
        let old_admin = keys[0];
```
nit: avoids Option handling

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+// === Getters ===
+
+/// `TypeName` of the protected root role.
+public fun protected_root<RootRole>(ac: &AccessControl<RootRole>): TypeName {
+    ac.protected_root
+}
+
+public fun max_default_admin_delay_ms(): u64 { MAX_DEFAULT_ADMIN_DELAY_MS }
+
+public fun default_admin_delay_ms<RootRole>(ac: &AccessControl<RootRole>): u64 {
```
0xNeshi, line 889: Missing docs on getters

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,572 @@
...
+}
+
+public fun pending_default_admin_new_admin<RootRole>(
+    ac: &AccessControl<RootRole>,
+): Option<address> {
+    if (ac.pending_default_admin.is_none()) return option::none();
+    option::some(ac.pending_default_admin.borrow().new_admin)
+}
+
+public fun pending_default_admin_execute_after_ms<RootRole>(
```
0xNeshi, line 943: Why not have a single `pending_default_admin: Option<PendingAdminTransfer>` getter?

0xNeshi (commented):
First pass

contracts/access/tests/access_control_tests.move
```diff
@@ -0,0 +1,1031 @@
...
+    let expected = access_control::test_new_role_admin_changed(
+        with_original_ids<RoleX>(),
+        with_original_ids<ACCESS_CONTROL_TESTS>(),
+        with_original_ids<AdminA>(),
+    );
+    assert_eq!(changed[0], expected);
+
+    // Grant AdminA to alice. alice can now grant RoleX (chain in effect),
+    // but the deployer (root) no longer can — RoleX's admin is now AdminA.
+    ac.grant_role<_, AdminA>(alice, scenario.ctx());
```
0xNeshi, line 486: Assert that root admin can no longer grant `RoleX` (since `AdminA` is now role admin)

contracts/access/tests/access_control_tests.move
```diff
@@ -0,0 +1,1031 @@
...
+
+    assert_eq!(ac.pending_default_admin_new_admin(), option::some(@0xC));
+    assert_eq!(ac.pending_default_admin_execute_after_ms(), option::some(50));
+
+    clock::destroy_for_testing(clk);
+    test_scenario::return_shared(ac);
+    scenario.end();
+}
+
+// === accept_default_admin_transfer ===
```
0xNeshi, line 836: Add a test case where `old_admin == new_admin`

contracts/access/tests/access_control_tests.move
```diff
@@ -0,0 +1,1031 @@
...
+    ac.begin_default_admin_transfer(@0xC, &clk, scenario.ctx());
+
+    assert_eq!(ac.pending_default_admin_new_admin(), option::some(@0xC));
+    assert_eq!(ac.pending_default_admin_execute_after_ms(), option::some(50));
+
+    clock::destroy_for_testing(clk);
+    test_scenario::return_shared(ac);
+    scenario.end();
+}
+
```
0xNeshi, line 835: Add test case asserting that there can be multiple holders of a non-root role

contracts/access/tests/access_control_tests.move
```diff
@@ -0,0 +1,1031 @@
...
+fun test_set_role_admin_rejects_root_subject() {
+    let deployer = @0xA;
+    let mut scenario = setup(deployer, 0);
+    let mut ac = take_ac(&scenario);
+    ac.set_role_admin<_, ACCESS_CONTROL_TESTS, AdminA>(scenario.ctx());
+    abort 0
+}
+
+#[test, expected_failure(abort_code = access_control::EUnauthorized)]
+fun test_set_role_admin_rejects_non_admin() {
```
0xNeshi, line 559: Add a similar test where a new role admin is set successfully, but then a non-role-admin user tries to invoke `set_role_admin` for the same `RoleX` (current test covers flow when `RoleX` admin is being added for the first time, the suggested test covers flow when the role admin already exists)

contracts/access/tests/access_control_tests.move
```diff
@@ -0,0 +1,1031 @@
...
+
+    clock::destroy_for_testing(clk);
+    test_scenario::return_shared(ac);
+    scenario.end();
+}
+
+// === cancel_default_admin_transfer ===
+
+#[test]
+fun test_cancel_admin_transfer_happy_path() {
```
0xNeshi, line 1034: Add a test case where a new admin transfer initiated after cancelling -> verifies that the cancel path leaves the registry usable for a fresh `begin_default_admin_transfer`

contracts/access/tests/access_control_tests.move
```diff
@@ -0,0 +1,1031 @@
...
+    scenario.end();
+}
+
+#[test, expected_failure(abort_code = access_control::ENoPendingAdminTransfer)]
+fun test_cancel_admin_transfer_rejects_no_pending() {
+    let deployer = @0xA;
+    let mut scenario = setup(deployer, 0);
+    let mut ac = take_ac(&scenario);
+    ac.cancel_default_admin_transfer(scenario.ctx());
+    abort 0
```
0xNeshi: can these `abort 0` be removed?

contracts/access/sources/access_control.move
ericnordelo: I don't think this should be the behavior, even if there is no one with the Role, that doesn't mean the role doesn't make sense semantically, and removing the role admin in this flow would go potentially unnoticed.

contracts/access/sources/access_control.move
ericnordelo: We could do it, but note that strucut members can't be directly accessed outside the defining module, so we would then need to add getters for each member anyway, just scoped in the struct instead of in AccessControl, and we need to add copy to each struct. I don't think there's a strong benefit on doing that if the only target is querying the members.

contracts/access/tests/access_control_tests.move
ericnordelo: The abort 0 pattern lets us avoid the cleanup we would need to add for not dropable variables. To remove it in test_accept_admin_renounce_rejects_pending_transfer we would need to manually return ac, clk, and then end the scenario, I vote we keep the abort 0 approach.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// timelocked transfer or renounce flow.
+#[error(code = 1)]
+const ECannotManageRootRole: vector<u8> =
+    "Root role can only be changed via the delayed transfer or renounce flow";
+
+/// `accept_default_admin_transfer`, `accept_default_admin_renounce`, or
+/// `cancel_default_admin_transfer` called with no pending action in the
+/// shared transfer-or-renounce slot.
+#[error(code = 2)]
+const ENoPendingAdminTransfer: vector<u8> = "No pending default admin transfer or renounce";
```
0xNeshi: Listing function names where the error is used, putting the list in the error doc comment itself is a smell to me.
If we start aborting with the error in another function, we have to _remember_ to add the name here.

Consider rewording to avoid listing specific functions where the error is used.

_NOTE: applies to all errors that list specific functions._

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// `is_one_time_witness` check confirms the value is genuine — this is what
+/// enforces Invariant 1 (one registry per module per publish). The
+/// transaction sender automatically becomes the root role holder.
+///
+/// The expected call site is the consumer module's `init` function, where
+/// the VM has already produced exactly one OTW value.
+///
+/// #### Parameters
+/// - `RootRole` (type): the module's OTW type; becomes the root role of the registry.
+/// - `otw`: a VM-issued One-Time Witness of type `RootRole`. Consumed by the call.
```
0xNeshi: Consider removing `RootRole (type)` item, as it might be redundant since it's documented in `otw` item.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// - `default_admin_delay_ms`: timelock (ms) that applies to root role transfer and renounce. Mutable post-creation via `begin_default_admin_delay_change`.
+/// - `ctx`: transaction context. The sender becomes the initial root role holder.
+///
+/// #### Returns
+/// The freshly-minted singleton `AccessControl<RootRole>` registry. The caller
+/// is responsible for sharing, embedding, or otherwise positioning it.
```
0xNeshi, line 297:
```suggestion
/// #### Returns
/// - The freshly-minted singleton `AccessControl<RootRole>` registry. The caller
/// is responsible for sharing, embedding, or otherwise positioning it.
```
nit: Convention is to add dash before the start of the `Returns` sentence.
I'd be in favor of changing this convention (in a following PR), but for now let's add it for consistency.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+}
+
+/// Initiate a renounce of the root role.
+///
+/// Caller must hold the root role. The renounce cannot be finalized until
+/// `default_admin_delay_ms` has elapsed, giving the protocol a cancel window
+/// before the registry becomes permanently unmanaged. An existing pending
+/// transfer or renounce is overwritten.
+///
+/// **Warning:** Once finalized via `accept_default_admin_renounce`, no one
```
```suggestion
/// #### Security Warning
///
/// - Once finalized via `accept_default_admin_renounce`, no one
```
nit: consistent with [current convention](https://github.com/OpenZeppelin/contracts-sui/blob/0a996c95e2a7de700e61a434186709cd04f101f3/contracts/access/sources/ownership_transfer/two_step.move#L242)

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+
+/// Grant role `Role` to `account`. Caller must hold the admin role of `Role`.
+/// `Role` must be defined in the same module as `RootRole` (Invariant 2).
+/// No-op if `account` already holds `Role`.
+///
+/// #### Parameters
+/// - `Role` (type): the role to grant. Must be a home-module role.
+/// - `ac`: the registry to mutate.
+/// - `account`: the address being granted `Role`.
+/// - `ctx`: transaction context. The sender must hold the admin role of `Role`.
```
```suggestion
/// - `ctx`: transaction context.
```
nit: potentially redundant, documented in [Aborts section](https://github.com/OpenZeppelin/contracts-sui/pull/313/changes#diff-a8d2bcee6cb0a9f5c79d7dabcedc47433f84979e57d3a518a9c07108f32947c5R424).

Similarly in other relevant functions.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+public struct PendingDelayChange has drop, store {
+    new_delay_ms: u64,
+    schedule_after_ms: u64,
+}
+
+// === Events ===
+
+/// Emitted when a role is granted to an account. Fired by `grant_role` (when
+/// `account` was not already a member), by `new` for the initial root role
+/// holder, and by `accept_default_admin_transfer` for the incoming new admin.
```
0xNeshi: Similar to errors, listing functions in which events are fired is a smell. Is there a benefit to this approach that makes the drawback(s) acceptable? If not, I'd remove function mentions

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+///
+/// The `role` field is a `TypeName` — it embeds the defining package address
+/// and module name of the role struct, so events from independently-published
+/// consumer protocols are distinguishable off-chain.
+public struct RoleGranted has copy, drop {
+    /// The role granted.
+    role: TypeName,
```
0xNeshi, line 219: Since each event field has its own docs, the comment part ```///The `role` field is a `TypeName`... ``` seems better suited to be placed above the `role` field itself.

Similar in other events.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+    previous_admin_role: TypeName,
+    /// The admin role in effect after the call.
+    new_admin_role: TypeName,
+}
+
+/// Emitted when a transfer of the root role is scheduled by
+/// `begin_default_admin_transfer`. Distinct from `DefaultAdminRenounceScheduled`
+/// so off-chain consumers can tell the two kinds of pending state apart
+/// without inspecting payload.
+public struct DefaultAdminTransferScheduled has copy, drop {
```
```suggestion
/// Emitted when a transfer of the root role is scheduled. Distinct from
/// `DefaultAdminRenounceScheduled` so off-chain consumers can tell
/// the two kinds of pending state apart without inspecting payload.
public struct DefaultAdminTransferScheduled has copy, drop {
```
Keeps clarity while removing function mention. Similar change can be done on other events.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// and the configured delay must have elapsed. Atomically revokes from the
+/// previous holder and grants to the caller.
+///
+/// #### Aborts
+/// - `ENoPendingAdminTransfer` if no pending action exists.
+/// - `ENotPendingTransfer` if the pending action is a renounce, not a transfer.
+/// - `ENotPendingAdmin` if the caller is not the scheduled new admin.
+/// - `EDelayNotElapsed` if the timelock has not elapsed.
+///
+/// #### Parameters
```
0xNeshi: `Parameters` section should go above `Aborts` section

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// root role and the configured delay must have elapsed. Removes the caller
+/// from the root role; emits `RoleRevoked` and clears the pending slot.
+///
+/// #### Aborts
+/// - `ENoPendingAdminTransfer` if no pending action exists.
+/// - `ENotPendingRenounce` if the pending action is a transfer, not a renounce.
+/// - `EUnauthorized` if the caller does not hold the root role.
+/// - `EDelayNotElapsed` if the timelock has not elapsed.
+///
+/// #### Parameters
```
0xNeshi: `Parameters` section should go above `Aborts` section

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// overwritten.
+///
+/// #### Parameters
+/// - `ac`: the registry to mutate.
+/// - `new_delay_ms`: the proposed new value of `default_admin_delay_ms`. Bounded by `MAX_DEFAULT_ADMIN_DELAY_MS`.
+/// - `clock`: current clock; used to compute the schedule timestamp.
+/// - `ctx`: transaction context. The sender must hold the root role.
+///
+/// #### Aborts
+/// - `EUnauthorized` if the caller does not hold the root role.
```
0xNeshi, line 754:
```suggestion
/// Caller must hold the root role. An existing pending delay change is
/// overwritten.
///
/// #### Parameters
/// - `ac`: the registry to mutate.
/// - `new_delay_ms`: the proposed new value of `default_admin_delay_ms`. Bounded by `MAX_DEFAULT_ADMIN_DELAY_MS`.
/// - `clock`: current clock; used to compute the schedule timestamp.
/// - `ctx`: transaction context.
///
/// #### Aborts
/// - `EUnauthorized` if the caller does not hold the root role.
```
"Caller must hold the root role" is mentioned 3 times, at least the `ctx`-adjacent mention is redundant. Maybe even the first mention is too.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// No authorization required — the schedule was committed by the root admin
+/// at `begin` time; `accept` is just the state transition. Anyone can call
+/// it (matches the OZ semantic where the new delay takes effect after the
+/// schedule passes regardless of caller).
+///
+/// #### Aborts
+/// - `ENoPendingDelayChange` if no pending change exists.
+/// - `EDelayNotElapsed` if `schedule_after_ms` has not been reached.
+///
+/// #### Parameters
```
0xNeshi: Parameters goes before Aborts

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// - `ac`: the registry to mutate.
+/// - `clock`: current clock; used to verify the schedule has passed.
+/// - `_ctx`: transaction context. Not used (no authorization required) — kept for API uniformity with other accept functions.
```
```suggestion
/// - `_ctx`: transaction context.
```
Redundant info, just clutters the docs

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// `Auth<Role>` produced is unforgeable in context: the only registry that can
+/// mint it is the unique one rooted at `Role`'s module.
+///
+/// #### Parameters
+/// - `Role` (type): the role for which to mint an auth witness. Must be a home-module role.
+/// - `ac`: the registry to query.
+/// - `ctx`: transaction context. The sender must currently hold `Role`.
+///
+/// #### Returns
+/// An `Auth<Role>` capability bound to `ctx.sender()`. Pass it by reference
```
```suggestion
/// #### Returns
/// - An `Auth<Role>` capability bound to `ctx.sender()`. Pass it by reference
```

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// - `ac`: the registry to query.
+///
+/// #### Returns
+/// The `TypeName` of the registry's root role, captured at construction
+/// time and immutable thereafter.
+public fun protected_root<RootRole>(ac: &AccessControl<RootRole>): TypeName {
+    ac.protected_root
+}
+
+/// Upper bound on `default_admin_delay_ms`. See `MAX_DEFAULT_ADMIN_DELAY_MS`.
```
```suggestion
/// Upper bound on `default_admin_delay_ms`.
```
Redundant info? Move doesn't support code navigation in IDEs, and also public docs link to a private constant.

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+const EZeroAddress: vector<u8> = "Cannot use the zero address as a role holder or transfer target";
+
+// === Constants ===
+
+/// Upper bound on `default_admin_delay_ms` — the configured timelock for
+/// root role transfer / renounce. Enforced by `new` and by
+/// `begin_default_admin_delay_change`.
+///
+/// Value: 60 days (≈ 2 calendar months) in milliseconds.
+const MAX_DEFAULT_ADMIN_DELAY_MS: u64 = 60 * 24 * 60 * 60 * 1_000;
```
0xNeshi, line 123:
```suggestion
/// Upper bound on `default_admin_delay_ms` — the configured timelock for
/// root role transfer / renounce.
///
/// Value: 60 days (≈ 2 calendar months) in milliseconds.
const MAX_DEFAULT_ADMIN_DELAY_MS: u64 = 60 * 24 * 60 * 60 * 1_000;
```
nit: potential smell

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+/// time and immutable thereafter.
+public fun protected_root<RootRole>(ac: &AccessControl<RootRole>): TypeName {
+    ac.protected_root
+}
+
+/// Upper bound on `default_admin_delay_ms`. See `MAX_DEFAULT_ADMIN_DELAY_MS`.
+///
+/// #### Returns
+/// The compile-time constant `MAX_DEFAULT_ADMIN_DELAY_MS` (in ms).
+public fun max_default_admin_delay_ms(): u64 { MAX_DEFAULT_ADMIN_DELAY_MS }
```
0xNeshi, line 883:
```suggestion
/// #### Returns
/// - Maximum delay for root role transfer / renounce (in ms).
public fun max_default_admin_delay_ms(): u64 { MAX_DEFAULT_ADMIN_DELAY_MS }
```
nit: clearer + omits mention of private constant

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+}
+
+// === Default Admin Delay Change Getters ===
+
+/// Maximum wait before a scheduled delay *increase* takes effect.
+/// See `MAX_DELAY_INCREASE_WAIT_MS` for the full rationale.
+///
+/// #### Returns
+/// The compile-time constant `MAX_DELAY_INCREASE_WAIT_MS` (in ms).
+public fun max_delay_increase_wait_ms(): u64 {
```
0xNeshi, line 953: See comments above for `max_default_admin_delay_ms`

contracts/access/sources/access_control.move
0xNeshi, line 219: Convention for events in existing contracts is that fields don't have their own docs. Should we change the convention (so that fields have docs, as in this PR)?

contracts/access/README.md
```diff
@@ -1,20 +1,42 @@
...
-Both modules accept any `T: key + store` object, but the primary use case is wrapping **capabilities** — admin caps, treasury caps, and similar authority tokens — where a misdirected transfer could be irreversible.
+Transfer-policy modules accept any `T: key + store` object, but the primary use case is wrapping **capabilities** — admin caps, treasury caps, and similar authority tokens — where a misdirected transfer could be irreversible. The access-control module gates protocol functions by typed role witnesses without per-call ID pinning.
 
 ---
 
 ## Module Snapshot
 
 | Module | Summary |
 |--------|---------|
+| `access_control` | Role-based access control with self-validating `Auth<R>` capabilities, OTW-pinned singleton registry, and timelocked root role transfer. |
```
0xNeshi:
```suggestion
| `access_control` | Role-based access control with self-validating `Auth<Role>` capabilities, OTW-pinned singleton registry, and timelocked root role transfer. |
```
Update all `Auth<R>` into `Auth<Role>` for consistency with code base

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1131 @@
...
+
+    let role_data = ac.roles.borrow_mut<_, RoleData>(role_name);
+    if (!role_data.members.contains(&account)) return;
+
+    role_data.members.remove(&account);
+    event::emit(RoleRevoked { role: role_name, account, sender: ctx.sender() });
+}
+
+/// Voluntarily relinquish role `Role`. `account` must equal `ctx.sender()`.
+/// No-op if the caller does not hold `Role`. **Blocked on the root role** —
```
0xNeshi, line 461:
```suggestion
/// No-op if the caller does not hold `Role`.  —
```
Redundant (already documented in `Aborts`)?

contracts/access/README.md
```diff
@@ -1,20 +1,42 @@
...
+
+1. **Deploy**
+   - In your module's `init`, call `access_control::new<MY_OTW>(otw, default_admin_delay_ms, ctx)` to mint the singleton `AccessControl<MY_OTW>`. The publishing transaction's sender automatically becomes the root role holder. Share or embed the registry as your protocol requires.
+2. **Wire roles**
+   - Define phantom marker types in the same module as your OTW (e.g. `public struct AdminRole {}`, `public struct OperatorRole {}`).
+   - Build the admin hierarchy with `set_role_admin<RootRole, Role, AdminRole>` — every fresh role's admin defaults to the root role; rewire as needed.
+   - Use `grant_role<RootRole, R>`, `revoke_role<RootRole, R>`, and `renounce_role<RootRole, R>` to manage membership. Granting / revoking is gated by holding `R`'s admin role; renounce is self-only.
+3. **Issue capabilities**
+   - `new_auth<RootRole, R>(&registry, ctx)` mints `Auth<R>` for `ctx.sender()` if they currently hold `R`.
```
0xNeshi:
```suggestion
   - `new_auth<RootRole, Role>(&registry, ctx)` mints `Auth<Role>` for `ctx.sender()` if they currently hold `Role`.
```

contracts/access/sources/access_control.move
ericnordelo: I removed the field docs instead, to avoid the noise, to many comments.

ericnordelo: Removed!

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1094 @@
...
+
+    event::emit(DefaultAdminDelayChangeScheduled { new_delay_ms, schedule_after_ms });
+}
+
+/// Apply a pending delay change once its schedule has elapsed.
+///
+/// No authorization required — the schedule was committed by the root admin
+/// at `begin` time; `accept` is just the state transition. Anyone can call
+/// it (matches the OZ semantic where the new delay takes effect after the
+/// schedule passes regardless of caller).
```
0xNeshi: Text in braces possibly redundant

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1094 @@
...
+/// No-op if the caller does not hold `Role`. **Blocked on the root role** —
+/// use `begin_default_admin_renounce` + `accept_default_admin_renounce`
+/// instead, so the protocol gets the configured timelock and a cancel window
+/// before the registry becomes unmanaged.
+///
+/// #### Parameters
+/// - `Role` (type): the role to relinquish. Must be a home-module role.
+/// - `ac`: the registry to mutate.
+/// - `account`: the address relinquishing `Role`. Must equal `ctx.sender()`. The explicit parameter mirrors `grant_role` / `revoke_role`
+/// for API symmetry with OZ AccessControl conventions; the self-only constraint is enforced at runtime via `ECannotRenounceForOtherAccount`.
```
```suggestion
/// - `account`: the address relinquishing `Role`.
```
Redundant, documented in `Aborts`

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1094 @@
...
+    role_data.members.remove(&account);
+    event::emit(RoleRevoked { role: role_name, account, sender: ctx.sender() });
+}
+
+/// Set the admin role of `Role` to `AdminRole`. Both must be home-module
+/// roles. Caller must hold the current admin role of `Role`.
+///
+/// #### Parameters
+/// - `Role` (type): the role whose admin is being changed. Must be a home-module role and not the root role.
+/// - `AdminRole` (type): the role granted authority over `Role` going forward. Must be a home-module role.
```
```suggestion
/// Set the admin role of `Role` to `AdminRole`. Both must be home-module
/// roles. Caller must hold the current admin role of `Role`.
///
/// #### Parameters
/// - `Role` (type): the role whose admin is being changed.
/// - `AdminRole` (type): the role granted authority over `Role` going forward.
```
redundant, already mentioned in 2. and 3. sentence.

Please double check other docs for redundancies.

0xNeshi (approved):
Excellent! 

Approving the implementation, going through the docs site

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1073 @@
...
+    assert!(ac.has_role_by_name(ac.protected_root, ctx.sender()), EUnauthorized);
+    assert!(new_delay_ms <= MAX_DEFAULT_ADMIN_DELAY_MS, EDelayTooLarge);
+
+    let current = ac.default_admin_delay_ms;
+    let wait = if (new_delay_ms > current) {
+        if (new_delay_ms < MAX_DELAY_INCREASE_WAIT_MS) {
+            new_delay_ms
+        } else {
+            MAX_DELAY_INCREASE_WAIT_MS
+        }
```
```suggestion
        new_delay_ms.min(MAX_DELAY_INCREASE_WAIT_MS)
```

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1073 @@
...
+    /// specific new admin or a renounce. See `PendingAdminTransfer`.
+    pending_default_admin: Option<PendingAdminTransfer>,
+    /// Pending change to `default_admin_delay_ms`. The change becomes
+    /// effective only after `accept_default_admin_delay_change` is called
+    /// past `schedule_after_ms`; an in-flight `pending_default_admin` is
+    /// unaffected (it locks in the delay it was scheduled under).
+    pending_default_admin_delay_change: Option<PendingDelayChange>,
+    /// Current minimum delay (ms) for root role transfer / renounce.
+    /// Mutable through the delayed change flow only.
+    default_admin_delay_ms: u64,
```
0xNeshi, line 179:
```suggestion
    default_admin_change_delay_ms: u64,
```
nit: just a clarity-improvement suggestion, you don't have to apply if you think the current is clear enough

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1073 @@
...
+///    which compares the package's *original* publish address — so role
+///    types introduced in later upgrades of the same package pass the check
+///    too. The home-module rule restricts roles to the same package, not to
+///    the same package version.
+///
+/// Together these guarantees make `Auth<Role>` a self-validating capability.
+/// `Role` can only live in the registry of its home module, that registry is
+/// unique per publish, and the only path to mint `Auth<Role>` is `new_auth`
+/// against that registry. Action functions can take `&Auth<Role>` directly with
+/// no body checks.
```
immrsd, line 30: I reckon the `with no body checks` part here is misleading. The action functions on the client side have to make the `ac.hasRole` call to ensure the role hasn't been revoked (a user still owns the `Auth<Role>` object after the revocation, but has already been removed from `members`)

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1073 @@
...
+    /// unaffected (it locks in the delay it was scheduled under).
+    pending_default_admin_delay_change: Option<PendingDelayChange>,
+    /// Current minimum delay (ms) for root role transfer / renounce.
+    /// Mutable through the delayed change flow only.
+    default_admin_delay_ms: u64,
+}
+
+/// Per-role state: current members and the role that administers it.
+public struct RoleData has store {
+    members: VecSet<address>,
```
immrsd, line 184: Have you considered the idea of using a data structure with constant read/check complexity?

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1073 @@
...
+///
+/// #### Parameters
+/// - `Role` (type): the role whose admin role is being looked up.
+/// - `ac`: the registry to query.
+///
+/// #### Returns
+/// - The `TypeName` of `Role`'s admin role. Defaults to the root role's
+/// `TypeName` for roles that have no entry yet.
+public fun get_role_admin<RootRole, Role>(ac: &AccessControl<RootRole>): TypeName {
+    ac.get_role_admin_name(with_original_ids<Role>())
```
immrsd, line 368: It might make sense to include `assert_home_module<RootRole, Role>();` check in the public read functions to make user-facing errors clearer. In this particular case `get_role_admin` called with a foreign role will return the admin for `RootRole` which doesn't make much sense

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1073 @@
...
+}
+
+/// Aborts with `EUnauthorized` if `account` does not hold role `Role`.
+///
+/// #### Parameters
+/// - `Role` (type): the role to assert membership of.
+/// - `ac`: the registry to query.
+/// - `account`: the address being checked.
```
immrsd: Missing function descripiton

contracts/access/sources/access_control.move
0xNeshi: Depends on what is the expected general size of `members`. If a dozen or so, then this should be fine.
I don't have the data, but if the size could grow to hundreds or thousands, then a constant read/check DS might make sense.

contracts/access/sources/access_control.move
ericnordelo: Not really: Auth<Role> has only drop, no key, no store. It cannot persist across PTBs. So there is no cross-block scenario. Revocation in block N means new_auth will fail in block N+1 and since PTBs are atomic there is no way of revoking while an Auth is active unless the admin is the Auth handler itself and is self attacking somehow.

contracts/access/sources/access_control.move
ericnordelo: Let's keep it as it is for consistency with Solidity and Cairo.

contracts/access/sources/access_control.move
ericnordelo: VecSet has O(n) contains / insert / remove. For typical RBAC patterns (admin role: 1 holder; operator role: 1–10 holders; custom roles: 1–50 holders), this should be cheaper than something like Table for the dynamic field storage overhead.

contracts/access/sources/access_control.move
ericnordelo: I've been thinking about this for a while, and for other functions doesn't really apply, but for get_role_admin specifically is worth considering as you mention. I'm leaning toward not modifying it, because is extra overhead on a function that is used in others like grant_role, where assert_home_module is already called. The benefit of returning with ForeignRole in get_admin_role is not big enough imo to add the redundancy. Curious if you guys think otherwise.

0xNeshi: Thinking about this further, the home module invariant is relevant only to the _modifying functions_ (`grant_role`, `revoke_role` etc.).

@immrsd is there a benefit to asserting the invariant on getters? If not, I actually think we should leave things as-is

contracts/access/sources/access_control.move
immrsd: makes sense, O(n) should be fine for most of use cases, more complex data structures are not justified

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1072 @@
...
+                members: vec_set::singleton(sender),
+                // Root role is self-administering.
+                admin_role: root_type,
+            },
+        );
+
+    event::emit(RoleGranted { role: root_type, account: sender, sender });
+
+    ac
+}
```
immrsd, line 336: I believe it makes sense to provide a public `share` function that just wraps the `transfer::share_object`. The rationale is better composability. It's explicitly recommended in [Move Best Practices](https://docs.sui.io/develop/write-move/move-best-practices#shared-objects) for libraries

If we make this change, we should also not forget to update the examples in the docs to use this helper instead of `transfer::public_share_object`

contracts/access/sources/access_control.move
```diff
@@ -0,0 +1,1072 @@
...
+/// Assert that `account` holds role `Role`. Aborts otherwise.
+///
+/// #### Parameters
+/// - `Role` (type): the role to assert membership of.
+/// - `ac`: the registry to query.
+/// - `account`: the address being checked.
+///
+/// #### Aborts
+/// - `EUnauthorized` if `account` does not hold role `Role`.
+public fun assert_role<RootRole, Role>(ac: &AccessControl<RootRole>, account: address) {
```
immrsd: `assert_has_role` might be a slightly better name for this function

contracts/access/sources/access_control.move
immrsd: I believe the behaviour of the function is not correct if it returns the admin for `RootRole` for an unknown foreign role. It is extremely easy to fix by adding the `assert_home_module` check.

Regarding the point "extra overhead on a function that is used in others like grant_role", it does not apply to `get_role_admin` function. The other functions (e.g. `grant_role`) call the internal `get_role_admin_name` helper function that does not have to exercise this assertion

contracts/access/sources/access_control.move
0xNeshi: [Move Best Practices](https://docs.sui.io/develop/write-move/move-best-practices#shared-objects) recommends this for `key`-only objects (for which `transfer::public_share_object` isn't available).

`AccessControl` [has `store` ability](https://github.com/OpenZeppelin/contracts-sui/pull/313/changes/BASE..c1de8088a9c55c2d7572a4a5f40e2f55bd9587a2#diff-a8d2bcee6cb0a9f5c79d7dabcedc47433f84979e57d3a518a9c07108f32947c5R155), so devs can always invoke `transfer::public_share_object`. Exposing a `share` function would serve no purpose in this case.

contracts/access/sources/access_control.move
ericnordelo: Great points, let's add it.

immrsd (approved):
LGTM!

bidzyyys (approved):
LGTM!

PR #315, opened 2026-05-08
feat: Implement Rate Limiter module

### LINKED ISSUE #193: [Feature]: Implement Rate Limiter module
Implement Rate Limiter module based on prior research in #136.

0xNeshi:
Resolves #193  

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog

## Release Notes

* **New Features**
  * New utility package introducing an embeddable rate-limiter with three configurable strategies: Token Bucket (periodic refills), Fixed Window (per-window capacity), and Cooldown (capacity with cooldown periods).

* **Documentation**
  * Added comprehensive package documentation including usage examples, API reference, and configuration guidelines.


contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,512 @@
...
+    }
+}
+
+// === Private Functions ===
+
+macro fun assert_bucket_config($capacity: u64, $refill_amount: u64, $refill_interval_ms: u64) {
+    assert!($capacity > 0, EZeroCapacity);
+    assert!($refill_amount > 0, EZeroRefillAmount);
+    assert!($refill_interval_ms > 0, EZeroRefillInterval);
+}
```
ericnordelo: I think this being a function should be inlined, so there's no real practical benefit (confirm with the Sui team please), and using a macro has the counter of working pretty bad with the coverage tools. I vote we use a function.

ericnordelo: Applies to the other two macros below.

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,512 @@
...
+    assert!($capacity > 0, EZeroCapacity);
+    assert!($window_ms > 0, EZeroWindow);
+}
+
+macro fun assert_cooldown_config($capacity: u64, $cooldown_ms: u64) {
+    assert!($capacity > 0, EZeroCapacity);
+    assert!($cooldown_ms > 0, EZeroCooldown);
+}
+
+fun bucket_accrue(
```
ericnordelo, line 646: This function, even while private, should be properly documented so readers can quickly understand the behavior and purpose.

contracts/utils/tests/rate_limiter_tests.move
```diff
@@ -0,0 +1,790 @@
...
+
+    // After 20 ms, two refill steps (2 * 5 = 10) are credited back.
+    clk.set_for_testing(20);
+    assert_eq!(rl.available(&clk), 20);
+
+    // Refill is capped at the configured capacity.
+    clk.set_for_testing(1000);
+    assert_eq!(rl.available(&clk), 30);
+
+    abort 100
```
ericnordelo: These tests fail locally for me, because the inner code is calling abort while the test doesn't expect it. If we want to keep the abort pattern to avoid cleaning up scenarios that's fine, but let's make sure that there are no collisions by using an even bigger number like 9999.

contracts/utils/tests/rate_limiter_tests.move
```diff
@@ -0,0 +1,790 @@
...
+}
+
+#[test]
```
ericnordelo, line 46: We need to modify the CI workflow to run this new package tests.

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,512 @@
...
+    if (elapsed_steps <= steps_to_full) {
+        let credit = elapsed_steps * refill_amount;
+        (last_refill_ms + elapsed_steps * refill_interval_ms, available + credit)
+    } else {
+        // Smallest step count that reaches capacity: `steps_to_full` if exactly divisible, `steps_to_full + 1`
+        // otherwise. `steps_to_full + 1` cannot overflow: `steps_to_full <= elapsed_steps - 1 < u64::MAX`.
+        let steps = if (headroom == steps_to_full * refill_amount) steps_to_full
+        else steps_to_full + 1;
+        (last_refill_ms + steps * refill_interval_ms, capacity)
+    }
```
ericnordelo, line 674: I found this issue:

bucket_accrue advances last_refill_ms only to the moment the bucket first reaches capacity, not to the latest elapsed refill boundary at now.

When the bucket has been full, or becomes full before now, any extra elapsed refill intervals remain stored as stale time. After a consume drains tokens, a second consume at the same timestamp can reuse that stale time to refill immediately.

Impact: a caller can consume more than the configured bucket capacity in a single timestamp. Worst case, if the bucket was already full and enough time has elapsed, they can consume roughly 2 * capacity back-to-back.

ericnordelo: Here is a prompt for the agent to confirm and explain the issue:

```
Audit `contracts/utils/sources/rate_limiter.move`, focusing on the `Bucket` variant and the private `bucket_accrue` function.

Please confirm whether the token bucket can allow more than `capacity` to be consumed at the same timestamp after the bucket has been full or reaches full before `now`.

Specifically inspect this logic:

- `elapsed_steps = (now - last_refill_ms) / refill_interval_ms`
- `headroom = capacity - available`
- `steps_to_full = headroom / refill_amount`
- fill branch:
  `let steps = if (headroom == steps_to_full * refill_amount) steps_to_full else steps_to_full + 1;`
  `(last_refill_ms + steps * refill_interval_ms, capacity)`

Use a concrete example:
- `capacity = 10`
- `refill_amount = 1`
- `refill_interval_ms = 100`
- `initial_available = 10`
- `last_refill_ms = 1_000_000`
- `now = 1_001_000`

Trace two consecutive `try_consume(10, clock)` calls at the same `now`.

Questions to answer:
1. What does `bucket_accrue` return on the first call?
2. What state does `try_consume` store after the first call?
3. What does `bucket_accrue` return on the second call?
4. Can both consumes succeed?
5. If yes, explain why this violates token bucket capacity semantics.
6. Also check the partially-full case, e.g. `available = 8`, to see whether stale refill intervals remain after the bucket first reaches capacity.
7. Suggest the minimal fix or invariant: elapsed refill intervals beyond capacity should be discarded, so `last_refill_ms` should advance to the latest completed refill boundary rather than only to the first step that reaches capacity.

```

ericnordelo (commented):
Finished the first pass @0xNeshi , thank you again for the patience.

contracts/utils/sources/rate_limiter.move
ericnordelo: Quick update, I found an issue with it that could have been easier to find with proper docs.

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,512 @@
...
+            if (new_available < amount) return false;
+            *last_refill_ms = new_last;
+            *available = new_available - amount;
+            true
+        },
+        RateLimiter::FixedWindow { capacity, window_ms, window_start_ms, available } => {
+            let steps = (now - *window_start_ms) / *window_ms;
+            if (steps != 0) {
+                *window_start_ms = *window_start_ms + steps * *window_ms;
+                *available = *capacity;
```
ericnordelo: Modifying this when the flow returns false is inconsistent with the Bucket behavior, should we avoid modifying on false? I Havn't thought to deep about it so maybe it makes sense as it is, just wanted to bring it into your attention.

contracts/utils/tests/rate_limiter_tests.move
0xNeshi: Rookie mistake on my part believing Claude Code that tests succeed

contracts/utils/sources/rate_limiter.move
0xNeshi: It doesn't really matter. Not sure how much gas cost we'd save otherwise, but it would make the code less readable either way. `Code Clarity > negligible gas saving`.

Note: Bucket flow can postpone updating `available` because it stores the calculated amounts in local, i.e. ephemeral vars (`new_last` and `new_available`).

contracts/utils/sources/rate_limiter.move
0xNeshi: - fix: https://github.com/OpenZeppelin/contracts-sui/pull/315/commits/a2a8b9a843bd63b3e1fe555bef319576991e08dc
- regression tests: https://github.com/OpenZeppelin/contracts-sui/pull/315/commits/56e05a82eab5d447e620da18cf3f7f496e5198a1

0xNeshi: Nice catch!

contracts/utils/sources/rate_limiter.move
bidzyyys, line 1: @0xNeshi curios to know why `utils` 👀 ? How about e.g. `contracts/defi`? I do not have strong opinion on that. 

@immrsd wdyt?

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+    /// Continuously refilling bucket. `available` accrues `refill_amount` every
+    /// `refill_interval_ms`, capped at `capacity`. Each `try_consume` draws `available` down.
+    Bucket {
+        capacity: u64,
+        refill_amount: u64,
+        refill_interval_ms: u64,
+        last_refill_ms: u64,
+        available: u64,
+    },
+    /// Up to `capacity` units per window of length `window_ms`, anchored at the limiter's
```
bidzyyys:
```suggestion

    /// Up to `capacity` units per window of length `window_ms`, anchored at the limiter's
```

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+    /// creation time. `available` resets to `capacity` when current time crosses into a
+    /// later window boundary.
+    FixedWindow {
+        capacity: u64,
+        window_ms: u64,
+        window_start_ms: u64,
+        available: u64,
+    },
+    /// Up to `capacity` units may be consumed before the limiter gates on `cooldown_ms`.
```
bidzyyys, line 131:
```suggestion

    /// Up to `capacity` units may be consumed before the limiter gates on `cooldown_ms`.
```

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+            );
+            if (new_available < amount) return false;
+            *last_refill_ms = new_last;
+            *available = new_available - amount;
+            true
+        },
+        RateLimiter::FixedWindow { capacity, window_ms, window_start_ms, available } => {
+            let steps = (now - *window_start_ms) / *window_ms;
+            if (steps != 0) {
+                *window_start_ms = *window_start_ms + steps * *window_ms;
```
bidzyyys: Is there any possibility to overflow on this operation?

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+/// - `self`: Limiter being charged.
+/// - `amount`: Units to consume.
+/// - `clock`: Reference to the Sui `Clock`, used to apply accrual / window rollover / cooldown release.
+///
+/// #### Returns
+/// - `true` if the consume succeeded, `false` if the limiter refused.
+///
+/// #### Aborts
+/// - `EInvalidAmount` if `amount == 0`.
+public fun try_consume(self: &mut RateLimiter, amount: u64, clock: &Clock): bool {
```
bidzyyys, line 330: Can we refactor this function? I mean
1. Have each Limiter type handled in a dedicated private function.
2. I can see many `if <cond> return false`. Is it possible to simplify it somehow?

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+            window_start_ms,
+            available,
+        } => {
+            assert_fixed_window_config(capacity, window_ms);
+            // roll forward under the OLD `window_ms` first; the new config takes
+            // effect from the rolled-forward anchor going forward. If a roll
+            // occurred, the fresh window starts with the NEW capacity available.
+            let steps = (now - *window_start_ms) / *window_field;
+            if (steps > 0) {
+                *window_start_ms = *window_start_ms + steps * *window_field;
```
bidzyyys: Any possibility to overflow here?

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+/// #### Parameters
+/// - `capacity`: Maximum token balance the bucket can hold.
+/// - `refill_amount`: Tokens credited per refill interval.
+/// - `refill_interval_ms`: Length of one refill interval, in milliseconds.
+///
+/// #### Aborts
+/// - `EZeroCapacity` if `capacity == 0`.
+/// - `EZeroRefillAmount` if `refill_amount == 0`.
+/// - `EZeroRefillInterval` if `refill_interval_ms == 0`.
+fun assert_bucket_config(capacity: u64, refill_amount: u64, refill_interval_ms: u64) {
```
bidzyyys: https://move-book.com/reference/abort-and-assert/clever-errors/#clever-errors-and-macros

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+/// Validate a `FixedWindow` configuration.
+///
+/// #### Parameters
+/// - `capacity`: Maximum units consumable per window.
+/// - `window_ms`: Length of one window, in milliseconds.
+///
+/// #### Aborts
+/// - `EZeroCapacity` if `capacity == 0`.
+/// - `EZeroWindow` if `window_ms == 0`.
+fun assert_fixed_window_config(capacity: u64, window_ms: u64) {
```
bidzyyys: https://move-book.com/reference/abort-and-assert/clever-errors/#clever-errors-and-macros

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+/// Validate a `Cooldown` configuration.
+///
+/// #### Parameters
+/// - `capacity`: Maximum units consumable per batch.
+/// - `cooldown_ms`: Wait between batches, in milliseconds.
+///
+/// #### Aborts
+/// - `EZeroCapacity` if `capacity == 0`.
+/// - `EZeroCooldown` if `cooldown_ms == 0`.
+fun assert_cooldown_config(capacity: u64, cooldown_ms: u64) {
```
bidzyyys: https://move-book.com/reference/abort-and-assert/clever-errors/#clever-errors-and-macros

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+    // than left as anchor drift that the next call would re-credit. The branch split keeps
+    // all intermediate u64 products bounded without relying on upper bounds on `capacity`
+    // or `refill_amount`:
+    //   * Under-fill: `elapsed_steps * refill_amount <= steps_to_full * refill_amount <= headroom <= capacity`,
+    //     so `available + credit <= capacity`. No overflow.
+    //   * Fill: write `capacity` directly. `elapsed_steps * refill_interval_ms <= now - last_refill_ms`,
+    //     so the new anchor stays `<= now`.
+    let headroom = capacity - available;
+    let steps_to_full = headroom / refill_amount;
+    let new_last = last_refill_ms + elapsed_steps * refill_interval_ms;
```
bidzyyys, line 665: Can it overflow?

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+    if (elapsed_steps <= steps_to_full) {
+        let credit = elapsed_steps * refill_amount;
```
bidzyyys, line 670: Can it overflow?

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,561 @@
...
+        },
+        RateLimiter::Cooldown { cooldown_ms: _, capacity, available, cooldown_end_ms } => {
+            if (*available > 0) *available
+            else if (now >= *cooldown_end_ms) *capacity
+            else 0
+        },
+    }
+}
+
+// === Reconfigure ===
```
bidzyyys: Shouldn't you mention here about access restriction to those functions? Like advertising our `RBAC` or any other access control mechanism?

contracts/utils/README.md
```diff
@@ -0,0 +1,63 @@
+# `openzeppelin_utils`
+
+Embeddable primitives for Sui smart contract development.
+
+## Module Snapshot
+
+| Module | Summary |
+|--------|---------|
+| `rate_limiter` | Three rate-limiting strategies behind one enum: token bucket, fixed window, and cooldown. |
```
bidzyyys, line 9: Explanation of rate-limiting strategies? Or at least some links to understand.

bidzyyys (commented):
Great job @0xNeshi;
Added first pass comments.

contracts/utils/sources/rate_limiter.move
0xNeshi: Because the rate limiter can be used outside of DeFi, e.g. in gaming: 
- character has health and mana
- health and mana regenerate over time (up to a cap)
- health and mana can be represented using a rate limiter.

contracts/utils/sources/rate_limiter.move
0xNeshi: I tried applying the change, but the Move's formatter reverts, i.e.[ removes the empty line](https://github.com/OpenZeppelin/contracts-sui/actions/runs/26094937425/job/76730084072?pr=315).
Leaving as-is.

contracts/utils/sources/rate_limiter.move
0xNeshi: Applied this, but @bidzyyys raised a [good point](https://github.com/OpenZeppelin/contracts-sui/pull/315#discussion_r3264770338) on why this should be kept macro.

Given that all these assertions are required in only 2 places, the best approach might be to just inline, i.e. duplicate, these assertions.

contracts/utils/sources/rate_limiter.move
0xNeshi: Originally applied due to https://github.com/OpenZeppelin/contracts-sui/pull/315#discussion_r3235842845 due to valid concerns about coverage.

However, given that all these assertions are required in only 2 places, the best approach might be to just inline, i.e. duplicate, these assertions.

contracts/utils/sources/rate_limiter.move
0xNeshi: https://github.com/OpenZeppelin/contracts-sui/pull/315#discussion_r3266020561

contracts/utils/sources/rate_limiter.move
0xNeshi: https://github.com/OpenZeppelin/contracts-sui/pull/315#discussion_r3266020561

contracts/utils/sources/rate_limiter.move
0xNeshi: > 1. Have each Limiter type handled in a dedicated private function.

You've read "Clean Code" too :smile: Jokes aside, the current function can be displayed on one screen, and analyzed in one go without hopping around the code base looking for each type's impl, making it easier to understand. I'd advise we keep it as-is.

> 2. I can see many `if <cond> return false`. Is it possible to simplify it somehow?

Only 3 instances, when we do the `amount` check. Forcing a single check on all 3 types would probably result in uglier code. Nevertheless, I don't think it's worth the effort anyway.

contracts/utils/sources/rate_limiter.move
0xNeshi: No - overflow is mathematically impossible here.

Here's the math:

- `steps = (now - window_start_ms) / window_ms` (integer division truncates)
- So `steps * window_ms ≤ now - window_start_ms`
- Therefore `window_start_ms + steps * window_ms ≤ now`

Since `now` is a `u64` from `clock.timestamp_ms()`, the result is bounded by a valid `u64`

contracts/utils/sources/rate_limiter.move
0xNeshi: https://github.com/OpenZeppelin/contracts-sui/pull/315#discussion_r3266502867

contracts/utils/sources/rate_limiter.move
0xNeshi: Similar as for https://github.com/OpenZeppelin/contracts-sui/pull/315#discussion_r3266502867

contracts/utils/sources/rate_limiter.move
0xNeshi: Since [steps_to_full * refill_amount <= headroom (u64)](https://github.com/OpenZeppelin/contracts-sui/pull/315/changes/BASE..a2a8b9a843bd63b3e1fe555bef319576991e08dc#diff-cee6db75e0500e9468f335bbf5a2b595141056abd901f1e85d5298c22f6fd8ebR553), and `elapsed_steps <= steps_to_full`, it must follow that `elapsed_steps * refill_amount` cannot overflow.

contracts/utils/sources/rate_limiter.move
0xNeshi: Add clarifying comments to all relevant code lines

contracts/utils/sources/rate_limiter.move
0xNeshi: The module intentionally does not concern itself with access controls, and leaves this to the integrator.

Additionally, the module doesn't depend on `openzeppelin_access`, and naming it would imply a recommended pairing that isn't real: integrators may equally well gate using other approaches.

I've added a single neutral note in the module-level "Operator responsibilities" section covering all `&mut` functions (not just `reconfigure_*`, since `consume_*` may also need gating depending on the call site) and explicitly listing `openzeppelin_access` alongside the other options without endorsing one, see https://github.com/OpenZeppelin/contracts-sui/pull/315/commits/cd03368e0804601c03bb1b6073d6630363fddf61.

contracts/utils/sources/rate_limiter.move
bidzyyys: All clear then!

bidzyyys (approved):
LGTM

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,544 @@
...
+            assert!(refill_interval_ms > 0, EZeroRefillInterval);
+
+            let (new_last, new_available) = bucket_accrue(
+                *last_refill_ms,
+                *available,
+                *cap_field,
+                *refill_amount_field,
+                *refill_interval_field,
+                now,
+            );
```
immrsd: If previous accrual that runs with old config accrues with `elapsed_steps == 0`, the `new_last` is the `last_refill_ms`. Then the next `try_consume` computes `elapsed_steps = (now - old_anchor) / NEW_interval_ms` 
So the OLD-config residual (up to OLD_refill_interval_ms - 1 ms) gets credited at the NEW rate

Concrete scenario: 
- capacity: 1000
- refill: 1/1_000 ms
- available: 0
At time = 999 (1 ms before the first OLD step would fire), admin reconfigures to refill at 1/10ms. The next `try_consume` at time=1_000 sees (1_000 − 0) / 10 = 100 elapsed steps and fills to `100`.

The statement "Accrues any tokens earned under the old rules first" is wrong in this specific case.

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,544 @@
...
+/// - `capacity`: Maximum units consumable per batch.
+/// - `cooldown_ms`: Wait, in milliseconds, between exhausting the batch and the next reset.
+///
+/// #### Returns
+/// - A new cooldown `RateLimiter` starting fully available.
+///
+/// #### Aborts
+/// - `EZeroCapacity` if `capacity == 0`.
+/// - `EZeroCooldown` if `cooldown_ms == 0`.
+public fun new_cooldown(capacity: u64, cooldown_ms: u64): RateLimiter {
```
immrsd: What do you think about adding `initial_available` parameter to `new_cooldown` and `new_fixed_window` (like already done for `new_bucket`)? It will make API more consistent and symmetric

Are there any strong reasons to omit this parameter for these particular rate-limiter configurations?

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,544 @@
...
+/// - `clock`: Reference to the Sui `Clock`, used to arm a fresh deadline if needed.
+///
+/// #### Aborts
+/// - `EWrongVariant` if the limiter is not currently a `Cooldown`.
+/// - `EZeroCapacity` if `capacity == 0`.
+/// - `EZeroCooldown` if `cooldown_ms == 0`.
+public fun reconfigure_cooldown(
+    self: &mut RateLimiter,
+    capacity: u64,
+    cooldown_ms: u64,
```
immrsd: IMO it makes sense to limit the cooldown_ms to a reasonably high value, i.e. 
`const MAX_COOLDOWN_MS: u64 = 100 * 365 * 24 * 60 * 60 * 1000; // 100 years`

If the cooldown is set to u64::max either by error or maliciously the next consume (of full available amount) will fail due to overflow here:
```
if (*available == 0 && now >= *cooldown_end_ms) {
    *cooldown_end_ms = now + cooldown_ms;
};
```

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,544 @@
...
+    /// is set to `now + cooldown_ms` - the absolute deadline at which the gate releases.
+    /// No further consume succeeds until `now >= cooldown_end_ms`, at which point
+    /// `available` resets to `capacity` and the next batch is granted. `cooldown_end_ms`
+    /// is taken into account only once the limiter has been drained and the gate is armed.
+    Cooldown {
+        cooldown_ms: u64,
+        capacity: u64,
+        available: u64,
+        cooldown_end_ms: u64,
+    },
```
immrsd, line 144: Shall we put `capacity` as the first field and `available` as the last? To make it more consistent with `Bucket` and `FixedWindow` fields

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,544 @@
...
+                *refill_interval_ms,
+                now,
+            );
+            accrued
+        },
+        RateLimiter::FixedWindow { capacity, window_ms, window_start_ms, available } => {
+            // A new window has begun once `window_ms` has elapsed since the current anchor.
+            if (now - *window_start_ms >= *window_ms) *capacity else *available
+        },
+        RateLimiter::Cooldown { cooldown_ms: _, capacity, available, cooldown_end_ms } => {
```
immrsd:
```suggestion
        RateLimiter::Cooldown { capacity, available, cooldown_end_ms, .. } => {
```

contracts/utils/README.md
```diff
@@ -0,0 +1,67 @@
...
+    // Cap at 1_000 tokens, refilling 100 every 6 s, starting full.
+    let withdraw_limiter = rate_limiter::new_bucket(1_000, 100, 6_000, 1_000, clock);
+    Vault { id: object::new(ctx), withdraw_limiter }
+}
+
+public fun withdraw(self: &mut Vault, amount: u64, clock: &Clock) {
+    self.withdraw_limiter.consume_or_abort(amount, clock);
+    // ... actual withdrawal ...
+}
+```
```
immrsd, line 81: Shall we include examples for `Cooldown` and `FixedWindow` and show that `available()` and `consume_or_abort()` calls are identical regardless of different rate-limiter types?

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,544 @@
...
+/// Initial available amount must not exceed capacity.
+#[error(code = 8)]
+const EInitialAboveCapacity: vector<u8> = "Initial available amount must not exceed capacity";
+
+// === Structs ===
+
+/// One embeddable limiter, three strategies. The variant is chosen at construction and can
+/// only be swapped by building a fresh `RateLimiter` and overwriting the field.
+///
+/// All variants store an `available` counter that starts equal to `capacity` and is
```
immrsd: `Bucket` actually starts with the provided `initial_available` value, not capacity

contracts/utils/README.md
```diff
@@ -0,0 +1,67 @@
...
+---
+
+## `rate_limiter` at a Glance
+
+`RateLimiter` is a `store + drop` value the integrator embeds as a field inside their own object.
+
+Three strategies share the same API:
+
+- **Bucket** ([Wikipedia](https://en.wikipedia.org/wiki/Token_bucket)) — a bucket holds up to `capacity` tokens and refills at a steady rate. Each consume drains tokens; an empty bucket denies the call. Permits short bursts up to `capacity` on top of a sustained average rate, which is what most APIs and on-chain throughput controls actually want.
+- **Fixed window** — time is partitioned into back-to-back windows of `window_ms`, anchored at limiter creation. Each window allows up to `capacity` units and resets to zero at the boundary. Cheap and easy to reason about; the known trade-off is that a caller can spend the full quota at the end of one window and the full quota at the start of the next, yielding a `2 * capacity` burst across the boundary. Pick it when the quota is the contract (e.g. "100 mints per hour") and the boundary burst is acceptable.
```
immrsd:
```suggestion
- **Fixed window** — time is partitioned into back-to-back windows of `window_ms`, anchored at limiter creation. Each window allows up to `capacity` units and resets to full `capacity` at the boundary. Cheap and easy to reason about; the known trade-off is that a caller can spend the full quota at the end of one window and the full quota at the start of the next, yielding a `2 * capacity` burst across the boundary. Pick it when the quota is the contract (e.g. "100 mints per hour") and the boundary burst is acceptable.
```
or "resets to zero usage"

immrsd (commented):
Looking good! Left a couple of suggestions

contracts/utils/sources/rate_limiter.move
0xNeshi: Isn't this the expected behavior?

My understanding of the `Accrues any tokens earned under the old rules first` statement is that `reconfigure_*` updates the rate limiter state according to the OLD rules, so that a `try_consume` in the same PTB or just at the same time would see the state as per the old rules first. On the next `Clock` tick, the `try_consume` would accrue tokens according to the NEW rules.
According to this understanding, at time 1000 the limiter _should_ return 100. 

What would you expect to be returned in this case?

immrsd: I think both approaches make sense. Based on the function's doc I'd expect the following behaviour when reconfiguring:
- accrue pending (0 since time elapsed < 1 step)
- move timestamp to `now`
- update config
- onwards accrue from `now`

0xNeshi: ~~The alternative would be to update the logic so that the following happens:~~
~~- at time of `reconfigure_*`, accrue tokens as per old rules~~
~~- update `last_refill_ms` to `now` (i.e. reset the refill counter)~~
~~- start accruing tokens as per new rules~~

~~This might be a valid alternative approach. In the edge case you raised, it would completely reset the refill counter and force users to wait a full refill cycle, but it might be a price worth paying.~~

~~Was this the behavior you expected?~~

EDIT: Just saw your comment. I agree that might be the intuitive approach.
Will update the logic and docs to this.

contracts/utils/sources/rate_limiter.move
0xNeshi: I don't think we should set an upper limit on the cooldown.

- **Setting it by error**: isn't really an argument imo, as any value can be set to `u64::max` by error, e.g. `refill_amount`, `refill_interval__ms` etc. It would mean we'd need to validate all possible parameters for errors. This is a user concern, not rate limiter policy.
- **Setting it maliciously**: this is a data structure devs integrate internally into their protocols, and they get to define who can set the cooldown. It is a safe assumption that devs wouldn't maliciously break their own protocols.

immrsd: The idea was that an attacker that got access to rate-limiter configuration can break the protocol until re-configuration process is completed. But I hear you, given that reconfiguration is always possible, I don't think it's a must to add an upper limit here

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,569 @@
...
+    assert!(cooldown_ms > 0, EZeroCooldown);
+    assert!(initial_available > 0, EZeroCooldownInitial);
+    assert!(initial_available <= capacity, EInitialAboveCapacity);
+
+    RateLimiter::Cooldown {
+        cooldown_ms,
+        capacity,
+        available: initial_available,
+        cooldown_end_ms: 0,
+    }
```
immrsd, line 291: nit
```suggestion
    RateLimiter::Cooldown {
        capacity,
        cooldown_ms,
        cooldown_end_ms: 0,
        available: initial_available,
    }
```

contracts/utils/Pub.local.toml
```diff
@@ -0,0 +1,15 @@
+# generated by Move
+# this file contains metadata from ephemeral publications
+# this file should not be committed to source control
```
immrsd: Should it be committed?

contracts/utils/sources/rate_limiter.move
0xNeshi: Actually, accruing value on failure _does_ matter for fixed window and cooldown. As noted in above message:

> Bucket flow can postpone updating `available` because it stores the calculated amounts in local, i.e. ephemeral vars (`new_last` and `new_available`).

In the current design the other two types do not do this, because otherwise it would be necessary to perform the same accrual calculations on each `available` call, too; current design makes this unnecessary, so `available` is slightly more performant.

Maybe there's a way to refactor the code so the behavior is consistent across all types, but this might not be worth the effort.

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,569 @@
...
+        },
+        _ => abort EWrongVariant,
+    }
+}
+
+/// Rewrite a `Cooldown` limiter's configuration in place.
+///
+/// `available` is clamped to the new capacity. If after the clamp `available == 0` - whether
+/// because a gate was already armed under the old config or because the clamp drained the
+/// batch - the cooldown deadline is reset to `now + cooldown_ms` under the new `cooldown_ms`.
```
immrsd: Is it actually possible for the clamp to drain the batch? The gate can be already armed under the old config, but that seems to be the only case. It seems we should remove the "or because the clamp drained the batch" part

contracts/utils/sources/rate_limiter.move
immrsd: This issue might be a bit more serious than it seems. The current behaviour for bucket is different from one of `FixedWindow` and `Cooldown`:
- for `Bucket`: failed `try_consume` (that returns `false`) does not change the persisted state
- for `FixedWindow` and `Cooldown`: even failed `try_consume` changes the state and accrues the pending value

At first glance, there is no apparent issue with that since the result behaviour of the component remains the same. But it actually affects how `reconfigure_cooldown` and `reconfigure_fixed_window` behave.

Cooldown example:
1. new_cooldown(capacity: 10, cooldown_ms: 100, initial_available: 10)
2. at t = 10: drain to 0 -> available = 0, gate at 110
3. at t = 200: try_consume(15) -> too much, still failed `try_consume` call resets available to 10
4. at t = 201: reconfigure with 500 ms cooldown -> `if (*available == 0)` arm at :509 does NOT fire -> the new gate with 500 ms is not armed -> available = 10
If there was no failed `try_consume` call (that should have no effect on the component's behaviour), available would still be 0 and once `if (*available == 0)` arm fires, the new cooldown is triggered

It should be pretty easy to fix leveraging local variables. Yes, it does make the code a little bit more complicated, but I believe it's justified given the impact on the actual logic:
```
// something like this
let usable = if (*available > 0) *available
  else if (now >= *cooldown_end_ms) *capacity
  else return false;
if (amount > usable) return false;
*available = usable - amount;
if (*available == 0) *cooldown_end_ms = now + *cooldown_ms;
true
```

immrsd: Same also applies to `FixedWindow`:
depending on if there was a failed `try_consume` call, `FixedWindow` can recharge either to `available` or to the new `capacity`

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,569 @@
...
+            window_ms: window_field,
+            window_start_ms,
+            available,
+        } => {
+            assert!(capacity > 0, EZeroCapacity);
+            assert!(window_ms > 0, EZeroWindow);
+
+            // Roll forward under the OLD `window_ms` first so the carried-over `available`
+            // reflects the old schedule; then re-anchor at `now` and install the new config.
+            // If a roll occurred, the new window starts with the NEW capacity available.
```
immrsd: nit: slightly more readable to declare the local var close to the line it's actually used in

upd: the rendered diff is weird. to clarify, the suggested change is to move the `let now = clock.timestamp_ms();` declaration into the match-arm block

```suggestion
    match (self) {
        RateLimiter::FixedWindow {
            capacity: cap_field,
            window_ms: window_field,
            window_start_ms,
            available,
        } => {
            assert!(capacity > 0, EZeroCapacity);
            assert!(window_ms > 0, EZeroWindow);

            // Roll forward under the OLD `window_ms` first so the carried-over `available`
            // reflects the old schedule; then re-anchor at `now` and install the new config.
            // If a roll occurred, the new window starts with the NEW capacity available.
            let now = clock.timestamp_ms();
```

contracts/utils/sources/rate_limiter.move
```diff
@@ -0,0 +1,569 @@
...
+            assert!(refill_interval_ms > 0, EZeroRefillInterval);
+
+            let (_, new_available) = bucket_accrue(
+                *last_refill_ms,
+                *available,
+                *cap_field,
+                *refill_amount_field,
+                *refill_interval_field,
+                now,
+            );
```
immrsd: nit: slightly more readable to declare the local var close to the line it's actually used in

update: the rendered diff is weird. to clarify, the suggested change is to move the `let now = clock.timestamp_ms();` declaration into the match-arm block

```suggestion
    match (self) {
        RateLimiter::Bucket {
            capacity: cap_field,
            refill_amount: refill_amount_field,
            refill_interval_ms: refill_interval_field,
            last_refill_ms,
            available,
        } => {
            assert!(capacity > 0, EZeroCapacity);
            assert!(refill_amount > 0, EZeroRefillAmount);
            assert!(refill_interval_ms > 0, EZeroRefillInterval);
            
            let now = clock.timestamp_ms();
            let (_, new_available) = bucket_accrue(
                *last_refill_ms,
                *available,
                *cap_field,
                *refill_amount_field,
                *refill_interval_field,
                now,
            );
```

contracts/utils/sources/rate_limiter.move
immrsd: the rendered diff is weird. to clarify, the suggested change is to move the `let now = clock.timestamp_ms();` declaration into the match-arm block

contracts/utils/sources/rate_limiter.move
0xNeshi: Nice catch!

The proposed fix doesn't solve the problem:
1. Drain at t=10 → `available=0, cooldown_end_ms=110`
2. No intervening `try_consume`
3. At t=201, `reconfigure_cooldown(500)` → `available` still `0` (clamp is a no-op), so the `if (*available == 0)` branch fires and arms a fresh 500 ms gate ending at `701` — even though the original gate released at 110.

That's the same bug, just triggered differently.

The actual solution would be to update `reconfigure_cooldown` to accrue-until-now under the old config first, then clamp. But will apply "no state update on failure" for consistency.

Reviewing fixed window next.

immrsd: Actually my concern was not about the new cooldown armed, but about the fact that a failed `try_consume` call actually changes how the component will behave on the next `reconfigure` call:
- `try_consume` modifies state despite the fact if fails
- `reconfigure` depends on the state and behaves differently

I suggest to just ensure that a failing `try_consume` call never modifies the storage by using local variables (as `Bucket` does). Already implemented for `Cooldown`, only `FixedWindow` left

0xNeshi: Ah, fair enough. Then it seems two bugs surfaced from your finding. 

Will make the following changes:
1. Ensure `try_consume` never modifies state on failure
2. Ensure `available` still returns up-to-date accrual values
3. Fix the `Cooldown` reset bug
4. Investigate whether the same bug occurs for fixed window

0xNeshi: https://github.com/OpenZeppelin/contracts-sui/pull/315/commits/717620dce50cc7dfe43bd7f9209e147de96d088c

immrsd (approved):
LGTM!

contracts/utils/sources/rate_limiter.move
0xNeshi: Reconfigure flows [will be removed](https://github.com/OpenZeppelin/contracts-sui/pull/326), making this thread deprecated. Resolving.

bidzyyys (approved):

PR #337, opened 2026-05-29
[N-03] Access Control Audit

ericnordelo:
## Summary by CodeRabbit

* **Breaking Changes**
  * Updated `renounce_role` signature: the account parameter has been removed. Callers must now rely on the transaction context instead of specifying an explicit account.

* **New Features**
  * Enhanced `new_with_admin` constructor to properly initialize with the provided initial administrator address and emit corresponding events.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/contracts-sui/pull/337?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/contracts/access/sources/access_control.move
+++ b/contracts/access/sources/access_control.move
@@ -344,7 +344,7 @@ public fun new_with_admin<RootRole: drop>(
     assert!(default_admin_delay_ms <= MAX_DEFAULT_ADMIN_DELAY_MS, EDelayTooLarge);
     assert!(initial_admin != @0x0, EZeroAddress);
 
-    let event_sender = ctx.sender();
+    let sender = ctx.sender();
     let root_type = with_original_ids<RootRole>();
 
     let mut ac = AccessControl<RootRole> {
@@ -367,7 +367,7 @@ public fun new_with_admin<RootRole: drop>(
             },
         );
 
-    event::emit(RoleGranted { role: root_type, account: initial_admin, sender: event_sender });
+    event::emit(RoleGranted { role: root_type, account: initial_admin, sender });
 
     ac
 }
--- a/contracts/access/sources/access_control.move
+++ b/contracts/access/sources/access_control.move
@@ -17,7 +17,7 @@
 ///    `set_role_admin`) plus the auth-mint path (`new_auth`) — checks that
 ///    the role type's home module matches the root role's home module.
 ///    Foreign role types are rejected at the boundary; they cannot be
-///    introduced into the bag. The check uses `type_name::with_original_ids`,
+///    introduced into the role map. The check uses `type_name::with_original_ids`,
 ///    which compares the package's *original* publish address — so role
 ///    types introduced in later upgrades to the same module in the same
 ///    package pass the check too. The home-module rule restricts roles to the
@@ -48,9 +48,9 @@
 module openzeppelin_access::access_control;
 
 use std::type_name::{TypeName, with_original_ids};
-use sui::bag::{Self, Bag};
 use sui::clock::Clock;
 use sui::event;
+use sui::table::{Self, Table};
 use sui::types;
 use sui::vec_set::{Self, VecSet};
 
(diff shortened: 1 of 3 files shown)


contracts/access/sources/access_control.move
```diff
@@ -79,43 +79,39 @@ const ENotPendingAdmin: vector<u8> = "Caller is not the pending admin";
...
 /// address has no signing key, so a role granted to it can never be exercised
 /// and a transfer scheduled to it can never be accepted.
-#[error(code = 12)]
+#[error(code = 11)]
 const EZeroAddress: vector<u8> = "Cannot use the zero address as a role holder or transfer target";
 
 /// The current root holder tried to schedule a no-op transfer to itself.
-#[error(code = 13)]
+#[error(code = 12)]
 const EDefaultAdminTransferToSelf: vector<u8> = "Cannot transfer default admin to self";
```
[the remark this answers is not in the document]
ericnordelo: The package hasn't being released yet, so we can break still.

bidzyyys (approved):
Looks good to me.

contracts/access/sources/access_control.move
```diff
@@ -309,16 +308,48 @@ public fun new<RootRole: drop>(
...
+///
+/// #### Returns
+/// - The freshly-minted singleton `AccessControl<RootRole>` registry. The caller
+/// is responsible for sharing, embedding, or otherwise positioning it.
+///
+/// #### Aborts
+/// - `ENotOneTimeWitness` if `otw` is not a true One-Time Witness.
+/// - `EDelayTooLarge` if `default_admin_delay_ms` exceeds `MAX_DEFAULT_ADMIN_DELAY_MS`.
+/// - `EZeroAddress` if `initial_admin` is `@0x0`.
+public fun new_with_admin<RootRole: drop>(
```
0xNeshi, line 348: Consider having just a single `new` function that takes in the `initial_admin` param. 
Having `new` as a wrapper around `new_with_admin` is sugar that doesn't meaningfully improve DevEx

contracts/access/sources/access_control.move
```diff
@@ -309,16 +308,48 @@ public fun new<RootRole: drop>(
...
+    initial_admin: address,
+    default_admin_delay_ms: u64,
+    ctx: &mut TxContext,
 ): AccessControl<RootRole> {
     assert!(types::is_one_time_witness(&otw), ENotOneTimeWitness);
     assert!(default_admin_delay_ms <= MAX_DEFAULT_ADMIN_DELAY_MS, EDelayTooLarge);
+    assert!(initial_admin != @0x0, EZeroAddress);
 
-    let sender = ctx.sender();
+    let event_sender = ctx.sender();
```
0xNeshi: nit: why the rename?

contracts/access/sources/access_control.move
```diff
@@ -330,13 +361,13 @@ public fun new<RootRole: drop>(
         .add(
             root_type,
             RoleData {
-                members: vec_set::singleton(sender),
+                members: vec_set::singleton(initial_admin),
```
0xNeshi: In the report comments, you wrote:
> Regarding 1: The current generic role storage keeps root behavior consistent with has_role, new_auth, events, and admin-role lookup. Moving root to Option&lt;address&gt; would reduce one internal singleton VecSet, but would add special cases throughout the module. 

But having root admin as a separate field actually simplifies the code, and makes it easier to reason about.
I opened a [PR](https://github.com/OpenZeppelin/contracts-sui/pull/340) into this branch for you to review

contracts/access/sources/access_control.move
ericnordelo: I did consider doing it that way. The rationale of adding the new function is explained in the comments, and I think it does (arguibly) improve DX. It is a matter of avoiding copy paste issues and enabling just the admin case from deployments from multisig.

See comments in lines 318, and 286.

0xNeshi: While reviewing I read the [comment](https://github.com/OpenZeppelin/contracts-sui/pull/337/changes/#diff-a8d2bcee6cb0a9f5c79d7dabcedc47433f84979e57d3a518a9c07108f32947c5R286-R287), with the relevant reason for having two `new` versions being:
> avoids copy-pasting an address into `init`.

I'm just doubting the slight DX benefit justifies the extra code/docs/tests.

You can resolve the comment if you think this should be left as-is.

contracts/access/sources/access_control.move
```diff
@@ -280,12 +276,15 @@ public struct DefaultAdminDelayChangeCancelled has copy, drop {}
...
-/// Create the singleton `AccessControl` for a module.
+/// Create the singleton `AccessControl` for a module, with the transaction
+/// sender as the initial root role holder.
 ///
 /// The caller passes their module's One-Time Witness as `otw`. The runtime
 /// `is_one_time_witness` check confirms the value is genuine — this is what
 /// enforces Invariant 1 (one registry per module, initialized at first
 /// publish). The transaction sender automatically becomes the root role holder.
+/// Prefer this constructor when the initial admin should be the caller; it
+/// avoids copy-pasting an address into `init`.
```
```suggestion
/// Prefer this constructor when the initial admin should be the caller; it
/// avoids copy-pasting an address into `new`.
```

ericnordelo: new should be called from init, I feel init is more accurate, but still the comment is somehow not entirely accurate, because you would not copy paste an address, but pass the sender from the init function into the new call.

ericnordelo: Will rephrase it.

contracts/access/sources/access_control.move
ericnordelo: Applied!

contracts/access/sources/access_control.move
0xNeshi: Unresolving, wasn't addressed

contracts/access/sources/access_control.move
ericnordelo: This was rephrased in the last iteration.

contracts/access/sources/access_control.move
ericnordelo: I rolled it back, then it was added again by the agent :*

contracts/access/sources/access_control.move
```diff
@@ -166,14 +166,20 @@ public struct AccessControl<phantom RootRole> has key, store {
     id: UID,
     /// Per-role membership and admin mapping. Keyed by `TypeName`.
     /// Entries are created lazily on first grant or `set_role_admin`.
-    roles: Bag,
+    /// Does NOT contain an entry for the root role; see `default_admin`.
```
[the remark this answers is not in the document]
ericnordelo: The contract is not published, update your memory

0xNeshi (approved):
:fire:

PR #338, opened 2026-05-29
[N-05] Access Control Audit

ericnordelo:
## Release Notes

* **Refactor**
  * Updated `default_admin_delay_ms` to require a Clock parameter for accurate delay information.
  * Removed `accept_default_admin_delay_change` function; admin delay changes now automatically materialize when elapsed.
  * Streamlined admin delay change flow for improved handling of pending delay updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/contracts-sui/pull/338?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/contracts/access/sources/access_control.move
+++ b/contracts/access/sources/access_control.move
@@ -181,11 +181,11 @@ public struct AccessControl<phantom RootRole> has key, store {
     /// Pending change to the default admin — either a transfer to a specific
     /// new admin or a renounce. See `PendingAdminTransfer`.
     pending_default_admin: Option<PendingAdminTransfer>,
-    /// Pending change to `default_admin_delay_ms`. The change becomes
-    /// effective once `schedule_after_ms` has elapsed. Until a delay-refreshing
-    /// entrypoint clears this slot, it remains stored as pending; an in-flight
-    /// `pending_default_admin` is unaffected (it locks in the delay it was
-    /// scheduled under).
+    /// Pending change to `default_admin_delay_ms`. The change becomes effective
+    /// once `schedule_after_ms` has elapsed. Until a delay-refreshing entrypoint
+    /// clears this slot, it may remain stored here even though public pending
+    /// getters no longer report it as pending. An in-flight `pending_default_admin`
+    /// is unaffected (it locks in the delay it was scheduled under).
     pending_default_admin_delay_change: Option<PendingDelayChange>,
     /// Current minimum delay (ms) for root role transfer / renounce.
     /// Mutable through the delayed change flow only.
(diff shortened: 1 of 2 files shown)


contracts/access/sources/access_control.move
```diff
@@ -1003,6 +1001,35 @@ public fun pending_default_admin_delay_change_schedule_after_ms<RootRole>(
 
 // === Internal Helpers ===
 
+/// Returns the delay that should be used at `clock.timestamp_ms()`, including
+/// elapsed pending changes that have not been materialized yet.
+fun effective_default_admin_delay_ms<RootRole>(ac: &AccessControl<RootRole>, clock: &Clock): u64 {
+    if (ac.pending_default_admin_delay_change.is_some()) {
+        let pending = ac.pending_default_admin_delay_change.borrow();
+        if (clock.timestamp_ms() >= pending.schedule_after_ms) {
+            return pending.new_delay_ms
```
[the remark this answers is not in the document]
ericnordelo: This is valid syntax

bidzyyys (approved):
LGTM

contracts/access/sources/access_control.move
```diff
@@ -787,53 +794,29 @@ public fun begin_default_admin_delay_change<RootRole>(
...
     ac: &mut AccessControl<RootRole>,
+    clock: &Clock,
     ctx: &mut TxContext,
 ) {
     assert!(ac.pending_default_admin_delay_change.is_some(), ENoPendingDelayChange);
     assert!(ac.has_role_by_name(ac.protected_root, ctx.sender()), EUnauthorized);
 
+    if (refresh_default_admin_delay(ac, clock)) return;
+
     let _ = ac.pending_default_admin_delay_change.extract();
```
```suggestion
    refresh_default_admin_delay(ac, clock);

    let _ = ac.pending_default_admin_delay_change.extract();
```
If I tried to cancel a delay change, and the tx succeeded, I would expect to see both:
- the old delay in place
- `DefaultAdminDelayChangeCancelled` emitted

The code here would surprise me on both accounts:
- after my cancellation, the _new_ delay would be in place
- no `DefaultAdminDelayChangeCancelled` was emitted

The correct behavior here should be to abort if a delay shouldn't be cancelled.

contracts/access/sources/access_control.move
```diff
@@ -1003,6 +1002,35 @@ public fun pending_default_admin_delay_change_schedule_after_ms<RootRole>(
...
+
+    let (new_delay_ms, schedule_after_ms) = {
+        let pending = ac.pending_default_admin_delay_change.borrow();
+        (pending.new_delay_ms, pending.schedule_after_ms)
+    };
+    if (clock.timestamp_ms() < schedule_after_ms) return false;
+
+    let _ = ac.pending_default_admin_delay_change.extract();
+    ac.default_admin_delay_ms = new_delay_ms;
+    true
```
0xNeshi, line 1087:
```suggestion
    let effective_delay_ms = ac.effective_default_admin_delay_ms(clock);

    if (effective_delay_ms == ac.default_admin_delay_ms) return false;

    let _ = ac.pending_default_admin_delay_change.extract();
    ac.default_admin_delay_ms = effective_delay_ms;
    true
```
Reuses existing effective delay logic

ericnordelo: I’d keep the current implementation here. `effective_default_admin_delay_ms` is a read helper, but `refresh_default_admin_delay` also has to clear elapsed pending storage.

The suggested equality check would miss elapsed no-op changes, where `new_delay_ms == default_admin_delay_ms`. In that case the effective delay equals the stored delay, but the pending slot still needs to be extracted. Otherwise stale internal pending state remains and a later schedule could treat it as an overwritten pending change.

contracts/access/sources/access_control.move
ericnordelo: Updated!

0xNeshi (approved):

PR #339, opened 2026-05-29
[N-01] Access Control Audit

ericnordelo:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed admin role transfer and renounce flows to properly refresh pending delays before scheduling new actions.

* **Improvements**
  * Enhanced role and admin change events with improved type parameterization for better tracking and filtering.
  * Admin delay getter now computes effective delays by accounting for pending changes and elapsed time.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/contracts-sui/pull/339?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/contracts/access/sources/access_control.move
+++ b/contracts/access/sources/access_control.move
@@ -181,11 +181,11 @@ public struct AccessControl<phantom RootRole> has key, store {
     /// Pending change to the default admin — either a transfer to a specific
     /// new admin or a renounce. See `PendingAdminTransfer`.
     pending_default_admin: Option<PendingAdminTransfer>,
-    /// Pending change to `default_admin_delay_ms`. The change becomes
-    /// effective once `schedule_after_ms` has elapsed. Until a delay-refreshing
-    /// entrypoint clears this slot, it remains stored as pending; an in-flight
-    /// `pending_default_admin` is unaffected (it locks in the delay it was
-    /// scheduled under).
+    /// Pending change to `default_admin_delay_ms`. The change becomes effective
+    /// once `schedule_after_ms` has elapsed. Until a delay-refreshing entrypoint
+    /// clears this slot, it may remain stored here even though public pending
+    /// getters no longer report it as pending. An in-flight `pending_default_admin`
+    /// is unaffected (it locks in the delay it was scheduled under).
     pending_default_admin_delay_change: Option<PendingDelayChange>,
     /// Current minimum delay (ms) for root role transfer / renounce.
     /// Mutable through the delayed change flow only.
@@ -833,26 +833,24 @@ public fun begin_default_admin_delay_change<RootRole>(
 
 /// Cancel an unelapsed pending delay change. Caller must hold the root role.
 ///
-/// If the pending delay change has already elapsed, this call applies it and
-/// clears the pending slot instead of cancelling it.
+/// If the pending delay change has already elapsed, this call aborts because
+/// there is no active pending change to cancel.
 ///
 /// #### Parameters
 /// - `ac`: the registry to mutate.
-/// - `clock`: current clock; used to materialize elapsed delay changes.
+/// - `clock`: current clock; used to detect elapsed delay changes.
 /// - `ctx`: transaction context.
 ///
 /// #### Aborts
-/// - `ENoPendingDelayChange` if no pending change exists.
+/// - `ENoPendingDelayChange` if no active pending change exists.
 /// - `EUnauthorized` if the caller does not hold the root role.
 public fun cancel_default_admin_delay_change<RootRole>(
     ac: &mut AccessControl<RootRole>,
     clock: &Clock,
     ctx: &mut TxContext,
 ) {
-    assert!(ac.pending_default_admin_delay_change.is_some(), ENoPendingDelayChange);
+    assert!(has_active_pending_default_admin_delay_change(ac, clock), ENoPendingDelayChange);
     assert!(ac.has_role_by_name(ac.protected_root, ctx.sender()), EUnauthorized);
-
-    if (refresh_default_admin_delay(ac, clock)) return;
 
     let _ = ac.pending_default_admin_delay_change.extract();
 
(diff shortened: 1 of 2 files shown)


bidzyyys (approved):
LGTM

contracts/access/sources/access_control.move
```diff
@@ -176,9 +176,10 @@ public struct AccessControl<phantom RootRole> has key, store {
     /// specific new admin or a renounce. See `PendingAdminTransfer`.
     pending_default_admin: Option<PendingAdminTransfer>,
     /// Pending change to `default_admin_delay_ms`. The change becomes
-    /// effective only after `accept_default_admin_delay_change` is called
-    /// past `schedule_after_ms`; an in-flight `pending_default_admin` is
-    /// unaffected (it locks in the delay it was scheduled under).
+    /// effective once `schedule_after_ms` has elapsed. Until a delay-refreshing
```
0xNeshi: Duplicate change from [[N-05] Access Control Audit](https://github.com/OpenZeppelin/contracts-sui/pull/338/changes#diff-a8d2bcee6cb0a9f5c79d7dabcedc47433f84979e57d3a518a9c07108f32947c5R179)

Intentional?

contracts/access/sources/access_control.move
```diff
@@ -200,50 +201,54 @@ public struct RoleData has store {
...
     new_admin: Option<address>,
     execute_after_ms: u64,
 }
 
-/// Snapshot of a pending change to `default_admin_delay_ms`. Becomes
-/// effective only after `accept_default_admin_delay_change` runs at or past
-/// `schedule_after_ms`.
+/// Snapshot of a pending change to `default_admin_delay_ms`. Becomes effective
+/// once `schedule_after_ms` has elapsed and is later cleared by a
+/// delay-refreshing entrypoint.
```
0xNeshi, line 218: Duplicate change from [[N-05] Access Control Audit](https://github.com/OpenZeppelin/contracts-sui/pull/338/changes#diff-a8d2bcee6cb0a9f5c79d7dabcedc47433f84979e57d3a518a9c07108f32947c5R210-R212)

Intentional?

contracts/access/sources/access_control.move
```diff
@@ -200,50 +201,54 @@ public struct RoleData has store {
 /// `new_admin = none` represents a pending renounce — the current root holder
 /// finalizes after the delay via `accept_default_admin_renounce`. The same
 /// pending slot is reused so a transfer and a renounce are mutually
-/// exclusive: scheduling either overwrites the other.
+/// exclusive: scheduling either cancels and overwrites the other.
```
0xNeshi, line 210: nit: I'm trying to understand the wording: 
1. There is no "OR part" related to the "either". Scheduling either
   1. cancels and overwrites the other
   2. or/otherwise it does what?
2. What is "the other" here?

Consider rewording to make it clearer

contracts/access/sources/access_control.move
```diff
@@ -200,50 +201,54 @@ public struct RoleData has store {
...
 }
 
 // === Events ===
 
+/// Access-control events are parameterized by `RootRole` so the event type
+/// identifies the registry without duplicating registry fields in every
+/// payload. Event attribution is available from the Sui event envelope
+/// (`SuiEvent.sender`) and is intentionally not duplicated in the Move event
+/// payloads.
+///
```
```suggestion
// Access-control events are parameterized by `RootRole` so the event type
// identifies the registry without duplicating registry fields in every
// payload. Event attribution is available from the Sui event envelope
// (`SuiEvent.sender`) and is intentionally not duplicated in the Move event
// payloads.

```
Comment refers to all events, not just `RoleGranted`

contracts/access/sources/access_control.move
```diff
@@ -573,15 +582,19 @@ public fun begin_default_admin_transfer<RootRole>(
     assert!(ac.has_role_by_name(ac.protected_root, ctx.sender()), EUnauthorized);
     assert!(new_admin != @0x0, EZeroAddress);
     assert!(new_admin != ctx.sender(), EDefaultAdminTransferToSelf);
+    refresh_default_admin_delay(ac, clock);
```
0xNeshi, line 615: Duplicate change from [[N-05] Access Control Audit](https://github.com/OpenZeppelin/contracts-sui/pull/338/changes#diff-a8d2bcee6cb0a9f5c79d7dabcedc47433f84979e57d3a518a9c07108f32947c5R578)

Intentional?

contracts/access/sources/access_control.move
```diff
@@ -658,15 +671,19 @@ public fun begin_default_admin_renounce<RootRole>(
     ctx: &mut TxContext,
 ) {
     assert!(ac.has_role_by_name(ac.protected_root, ctx.sender()), EUnauthorized);
+    refresh_default_admin_delay(ac, clock);
 
     let execute_after_ms = clock.timestamp_ms() + ac.default_admin_delay_ms;
+    if (ac.pending_default_admin.is_some()) {
+        event::emit(DefaultAdminTransferCancelled<RootRole> {});
```
0xNeshi: Beginning renouncing the admin emits a corresponding `DefaultAdminRenounceScheduled`, clearly separating this flow from admin transfers. 
However, cancelling the revocation emits `DefaultAdminTransferCancelled`, confusing the two flows.
Consider adding a separate `DefaultAdminRenounceCancelled` event

contracts/access/sources/access_control.move
```diff
@@ -778,65 +797,44 @@ public fun begin_default_admin_delay_change<RootRole>(
...
 ///
-/// #### Parameters
-/// - `ac`: the registry to mutate.
-/// - `clock`: current clock; used to verify the schedule has passed.
-/// - `_ctx`: transaction context.
-///
-/// #### Aborts
-/// - `ENoPendingDelayChange` if no pending change exists.
-/// - `EDelayNotElapsed` if `schedule_after_ms` has not been reached.
-public fun accept_default_admin_delay_change<RootRole>(
```
0xNeshi: There must have been a mixup between this PR and [[N-05] Access Control Audit](https://github.com/OpenZeppelin/contracts-sui/pull/338/changes#diff-a8d2bcee6cb0a9f5c79d7dabcedc47433f84979e57d3a518a9c07108f32947c5L804) changes

contracts/access/sources/access_control.move
```diff
@@ -200,50 +201,54 @@ public struct RoleData has store {
...
 ///
 /// Role identifiers are `TypeName`s — they embed the defining package address
 /// and module name of the role struct, so events from independently-published
 /// consumer protocols are distinguishable off-chain.
-public struct RoleGranted has copy, drop {
+public struct RoleGranted<phantom RootRole> has copy, drop {
     role: TypeName,
     account: address,
-    sender: address,
 }
```
0xNeshi, line 240: The audit report commended including `sender` in `RoleGranted/RoleRevoked`, and the benefit of having it is that off-chain monitors can immediately "see" who approved granting the role. Since there can be many `Role` admins, it is useful to know off-chain who exactly granted/revoked the role.

The audit suggested the same `sender` is added to admin-change-related events. Despite there only being a single default admin, the reason adding the `sender` field is beneficial is both alignment with `RoleX` events, and that off-chain monitors can immediately read current default admin's address from the event.

contracts/access/sources/access_control.move
ericnordelo: Yes

contracts/access/sources/access_control.move
ericnordelo: Yes

contracts/access/sources/access_control.move
ericnordelo: either doesn't need the OR on this context, is referring to either of the two flows mentioned in the previous sentence: transfer and renounce. Scheduling either of those, overwrites the other.

ericnordelo: For the sake of timelines, leaving this nit for future iterations including the cancel addition which I think is clear enough. Note that updating this comment means that I also have to go to the docsite API reference for consistency, and those shouldn't be blockers for us to release this feature.

contracts/access/sources/access_control.move
ericnordelo: Intentional

contracts/access/sources/access_control.move
ericnordelo: On Sui (different to the EVM), all the emitted events carry metadata by default including the sender, and JSON-RPC EventFilter and GraphQL EventFilter both filter natively on this sender, but not in payload which requires extra indexing.

Per my investigation is actually common not to include this on protocols on Sui (with a couple of exceptions of course)

Screenshot 2026-06-01 at 10 52 09

I think we should remove it, not make it the norm.

contracts/access/sources/access_control.move
immrsd: Agree with @0xNeshi here, it might be better to emit `DefaultAdminRenounceCancelled`/`DefaultAdminTransferCancelled` depending if the `pending_default_admin.new_admin` is zero or not

And it makes sense to introduce an internal helper since the same logic should live in both `begin_default_admin_renounce` and `begin_default_admin_transfer`

immrsd (approved):
LGTM! Left just a single comment regarding the cancellation events

contracts/access/sources/access_control.move
0xNeshi: Makes sense!

contracts/access/sources/access_control.move
```diff
@@ -200,50 +201,54 @@ public struct RoleData has store {
...
+/// (`SuiEvent.sender`) and is intentionally not duplicated in the Move event
+/// payloads.
+///
 /// Emitted when a role is granted to an account.
 ///
 /// Role identifiers are `TypeName`s — they embed the defining package address
 /// and module name of the role struct, so events from independently-published
 /// consumer protocols are distinguishable off-chain.
-public struct RoleGranted has copy, drop {
+public struct RoleGranted<phantom RootRole> has copy, drop {
```
0xNeshi, line 237: What is the benefit of marking the `RootRole` on a role granted event that was potentially done by a non-root admin?

ericnordelo: Enables off-chain queries like "Give me all the RoleGranted events in the last 2 weeks for this specific AccessControl registry". The key part is being able to query events for the specific registry.

contracts/access/sources/access_control.move
ericnordelo: Applied!

contracts/access/sources/access_control.move
0xNeshi: > for this specific AccessControl registry

With or without this phantom type, you'd need the `AccessControl.id` to query this.

The same `RootRole` can theoretically be used for multiple `AccessControl` instances.

ericnordelo: Not really, the new function consumes the OTW as it is passed by value. One of the invariants is that there can only be on AccessControl per module.

0xNeshi: Ok, I can see how someone might query "give me events for an AccessControl registry with this RootRole"

0xNeshi (approved):

0xNeshi (approved):

PR #345, opened 2026-06-02
Gaussian math: `cdf` function

### LINKED ISSUE #96: [Feature]: `cdf` for fixed-point math library

Follow the [reference](https://paragraph.com/@evandekim/on-chain-atomic-gaussian-math).

immrsd:
## What landed

`cdf` is defined as a regular function on both fixed-point types — `sd29x9_base::cdf(z: SD29x9): SD29x9` and `ud30x9_base::cdf(z: UD30x9): UD30x9`, each returning the same type as its input. Both compute Φ(z) as a **rational `N(|z|)/D(|z|)`** (two degree-9 polynomials), evaluated via Horner on a **sign-magnitude `u256` accumulator at WAD (10¹⁸)**, then divided back to the type's `10⁹` scale in one rounding step. Coefficients were fit **offline** (AAA, via `scipy`+`mpmath`) and committed as Move constants. New: `sources/internal/gaussian.move` (algorithm + Horner helpers) and `sources/internal/cdf_coefficients.move` (generated). Pure function — no objects, clock, randomness, or state.

## Reading order (~20 min)

1. `sd29x9_base::cdf` and `ud30x9_base::cdf` — the public entry points; both delegate to the same helper.
2. `internal/gaussian.move` — sign-magnitude `u256` arithmetic, `horner_eval!` macro, and the self-contained `cdf_nonneg_raw` central-domain evaluator (handles saturation and `Φ(0)` shortcut internally).
3. `internal/cdf_coefficients.move` — generated; skim only (trust comes from §5, not eyeballing).
4. `sd29x9_base.move` private `decompose` + new `is_negative` — the only edits to existing code on the SD29x9 side.

## Background: what Φ is, and why it can't be built like `ln`/`exp`

**What `cdf` computes.** Φ(z) is the standard-normal CDF: the probability that a
normally-distributed value lands ≤ z — the area under the bell curve left of z.
It underpins options pricing, risk models, and sampling. (`pdf` is the curve's
height; `inverse_cdf`/PPF is Φ run backwards.)

**Why it's hard on-chain.** Φ has no elementary formula — it's defined by an
integral with no closed form. The textbook ways to get it use `exp(−z²/2)` or the
error function `erf`, but Move has **no floating point and no transcendentals**,
only integer arithmetic on our fixed-point types.

**Why it's unlike functions the package already has.** `ln`, `exp`, `sqrt`,
`pow` each have a *self-contained on-chain recipe*: shrink the input to a small
range with an identity (e.g. `ln(2^k·m) = k·ln2 + ln(m)`), then run a few Newton
steps or a short series — the answer is computed straight from the input every
time. Φ has **no such trick**: no range-reduction identity, no series that's both
accurate and cheap on-chain. Doing it directly isn't impossible, just
impractical — slow and inaccurate in fixed point.

**The article's idea** (Evan Kim, "On-Chain Atomic Gaussian Math"). Don't compute
Φ — *approximate the whole curve* with a **rational function** `N(z)/D(z)` (one
polynomial divided by another). A good rational matches Φ to ~`10⁻⁹` using only
multiplies and one divide — no `exp`, no `erf`. Finding the best coefficients is
heavy numerical work, so the work is split:

- **Offline (Python), once:** an algorithm called **AAA** (Adaptive
  Antoulas–Anderson, 2018) is fed high-precision samples of Φ and automatically
  produces a near-optimal rational fit — curve-fitting that needs no hand-tuning.
  The coefficients are validated against a 100-digit reference and baked into the
  Move source as integer constants.
- **On-chain, every call:** just evaluate the two polynomials (Horner's method)
  and divide. Tiny, deterministic, constant gas.

**This is why the Python toolkit exists.** AAA fitting, 100-digit reference
values, and error validation cannot run on-chain (no floats, no arbitrary
precision, far too expensive). On-chain we keep only the cheap half — evaluating
fixed coefficients. The same toolkit will later produce `pdf` and `inverse_cdf`.

**Why this is the right call here:**
- **No transcendentals** → no new dependency, deterministic, cheap, constant gas.
- **Near-optimal at low degree** → hits the `5×10⁻⁹` target at degree 9.
- **Even symmetry is free** → fit only `[0, 6.3]`, reflect via Φ(−z) = 1 − Φ(z).
- **Proven** → matches a deployed reference; we re-derived our own coefficients
  (didn't copy theirs) and beat its accuracy (`~7×10⁻¹⁰`). Alternatives (`erf`,
  plain series, lookup tables) are rejected with reasons in `01-research`.

### `sd29x9_base::cdf` — the signed entry point

```move
public fun cdf(z: SD29x9): SD29x9 {
    let Components { mag, neg } = decompose(z.unwrap());
    let phi = cdf_nonneg_raw(mag as u128);
    let raw = if (neg) {
        assert!(phi >= HALF_RAW, EInternalNegSubUnderflow);
        common::scale!() - phi      // reflect: 1 − Φ(|z|)
    } else {
        phi
    };
    sd29x9::wrap(raw, false)
}
```

- **`decompose`, not `abs()`.** `abs()`/`negate()` abort on `min()`; `cdf` must
  be total. `sd29x9_base::decompose` (private, same module) is total and returns
  `Components { neg, mag }` which destructures directly — no accessor functions
  needed because `Components` is defined in this same module.
- **Saturation lives inside the helper.** `gaussian::cdf_nonneg_raw` short-circuits
  to `ONE_RAW` for `z_raw ≥ MAX_Z_RAW`, so neither base module replicates the guard.
  For negative inputs, the sign-flip `common::scale!() - phi` turns the saturated
  upper-bound into `0`.
- **Reflect by subtraction, asserted.** `common::scale!() - phi` can't underflow
  because Φ(|z|) ≥ 0.5 — and that's checked live, guarding against future regressions.
- **Same-type return.** The probability `[0, 1]` fits naturally in non-negative
  `SD29x9`; the package convention is "return the input type" (cf. `abs(SD29x9): SD29x9`).

### `ud30x9_base::cdf` — the unsigned entry point

```move
public fun cdf(z: UD30x9): UD30x9 {
    wrap(cdf_nonneg_raw(z.unwrap()))
}
```

One line. UD30x9 is inherently non-negative, so no decomposition, no sign-flip.
Both type paths route through the same `gaussian::cdf_nonneg_raw`, guaranteeing
bit-for-bit equivalence on positives.

### `cdf_nonneg_raw` (L123)

```move
if (mag == 0) return HALF_RAW;                       // Φ(0) bit-exact (0.5)
let z_wad = (mag as u256) * common::scale_u256!();   // 10^9 → 10^18 (WAD)
// bind coeff vectors to locals once, then:
let n = horner_eval!(z_signed, num.length(), |i| (num_mags[i], num_negs[i]));
let d = horner_eval!(z_signed, den.length(), |i| (den_mags[i], den_negs[i]));
assert!(!horner::is_neg(&n), EInternalNumNegative);                       // INV-8
assert!(!horner::is_neg(&d) && horner::mag(&d) > 0, EInternalDenNonPositive); // INV-7
let phi = horner::mul_div_nearest_u256(mag(&n), 10^9, mag(&d));          // N·10^9/D
if (phi > ONE_RAW) ONE_RAW else phi                                      // overshoot clamp
```

- **WAD internally, cast once.** Running at 10¹⁸ preserves the `~3×10⁻⁹`
  precision down to the 10⁹ output; the single divide is the only down-cast.
- **Φ(0) special case** → exact `0.5`, removes any sign-flip asymmetry at 0.
- **Bind coeff vectors once** so the Horner loop indexes locals, not module
  constants (this is why accessors return whole vectors — deviation D1).
- **Sign asserts = integrity checks** on the generated file: a corrupted
  coefficient that flips a sign on `[0, 6.3]` aborts here, before any consumer.
- **Overshoot clamp**: nearest rounding can yield `10⁹+1` near z=6.3; clamp keeps
  it a valid probability.

**Watch:** `internal/gaussian.move`'s private `MAX_Z_RAW` and
`internal/cdf_coefficients.move`'s `MAX_Z_WAD` are two sources of truth for the
same `6.3` threshold (one at UD30x9 scale, one at WAD). Consistent today, worth
unifying to prevent drift.

## `internal/gaussian.move`

- **`SignedScaled256 { mag: u256, neg: bool }`** — Move has no signed ints, and
  intermediates exceed `u128`, so sign-magnitude over `u256` is the simplest
  correct encoding (canonical zero = `(0, false)`). Abilities `copy, drop` only:
  not storable. *Declared `public` (D2): Move 2024 forbids `public(package)` on
  struct decls; package-internality is enforced by the abilities + package-only
  constructors.*
- **`signed_add`** (L81) / **`signed_mul_wad`** (L99): textbook sign-magnitude.
  `mul_wad` is a raw `(a*b)/WAD` (not a checked `mul_div`) — safe only because
  magnitudes are bounded (see budgets below). Flag if you want a checked multiply.
- **`mul_div_nearest_u256`** (L117): `round(a·b/d)` half-up; inlined to avoid a
  `math/core` dependency (D6).
- **`horner_eval!` macro** (L145): takes a coefficient-accessor lambda so `pdf`
  and inverse-CDF reuse it with their own tables, and so the caller can bind
  vectors to locals. Coefficients pulled in **ascending power order**. The
  empty-polynomial abort sits in a helper the macro calls (D3, macro scoping).

## Budgets to sanity-check

- **Overflow (INV-10):** peak `acc·z` ≈ `1.4×10³⁸`, peak `N·10⁹` ≈ `1.4×10²⁹`,
  `u256::MAX` ≈ `1.16×10⁷⁷`. ~39 orders of headroom — **but the bound depends on
  `|z| ≤ 6.3`**, which the saturation guard enforces. Break the guard, break the bound.
- **Error (INV-17):** contract ≤ `5×10⁻⁹` (5 ULP). Committed coefficients:
  `6.78×10⁻¹⁰` ≈ 1 ULP. ~7× headroom.

## Trusting the coefficients (without reading the numbers)

Generated, so the question is "do I trust the pipeline":

- **`derive.py`** (local, `scipy`+`mpmath`): sweeps AAA degree, picks the smallest
  meeting `5×10⁻⁹`, normalizes `D(0)=1`, checks `N≥0`/`D>0` and `coeff·WAD < 2^128`.
- **`validate.py`** (the **CI gate**, `scipy`+`numpy` only): parses the committed
  `.move` file and re-runs the CDF in Python with **integer math mirroring the
  Move code exactly** (`cdf_simulate`), asserting ≤ 5 ULP vs `scipy.stats.norm.cdf`.

So CI never re-runs AAA — it just confirms the committed numbers still meet the
contract against an independent oracle. `cdf_simulate` doubles as an executable
spec of `gaussian::cdf_nonneg_raw`. Coefficients are re-derived by us (not lifted);
the `AUTO-GENERATED` banner names the source scripts.

## Likely questions

- **Two scales?** WAD internally preserves precision; cast to `10⁹` once at the end.
- **Why does `SD29x9::cdf` return `SD29x9` instead of `UD30x9`?** Package convention: every function returns the type it was called on. Mirrors `abs(SD29x9): SD29x9` (a non-negative SD29x9). Consumers needing UD30x9 can cast at the call site.
- **Why does `cdf` live in `sd29x9_base.move` / `ud30x9_base.move` instead of `gaussian/`?** Same convention: per-type public API lives in the type's base module, alongside `add`, `abs`, `pow`, etc. `internal/gaussian.move` holds shared algorithm helpers that the base modules consume.
- **Why `decompose` not `abs()`?** `abs()` aborts on `min()`; `cdf` must be total. `decompose` works directly on the raw `u128` bit pattern, so it's defined for every `SD29x9` including `min()`.
- **Raw multiply overflow?** No, magnitudes are bounded by the saturation guard inside `cdf_nonneg_raw` (not a runtime check — fair to ask for `mul_div` as defense-in-depth).
- **Always a valid probability?** Yes: exact saturation + overshoot clamp + underflow assert.
- **State / randomness / objects?** None. PTB-safe.
- **Locks us in for `pdf`/`inverse_cdf`?** Opposite — `internal/gaussian.move`'s Horner machinery is generic on purpose; both will reuse it, and each will add its own `*_coefficients.move` next to `cdf_coefficients.move`.

## Summary by CodeRabbit

* **New Features**
  * Added standard-normal CDF (`cdf`) for fixed-point Gaussian math types (`SD29x9.cdf()` and `UD30x9.cdf()`)
  * Added `SD29x9.is_negative()` predicate
* **Documentation**
  * Expanded CDF README with semantics, accuracy, saturation behavior, and determinism notes
  * Added documentation for the offline Gaussian CDF code-generation toolkit
* **Tests / CI**
  * Added comprehensive Horner and CDF correctness tests, including deterministic generated test vectors
  * Introduced an offline codegen validation workflow to detect drift and run codegen unit tests
* **Chores**
  * Added tooling/formatting support for the Gaussian code-generation pipeline


.github/workflows/test.yml
```diff
@@ -132,3 +132,28 @@ jobs:
...
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  codegen:
+    name: Codegen (validate + drift + tests)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
immrsd: Skipping for now — none of our other workflow checkouts set persist-credentials: false either, so I'd rather do this repo-wide in a dedicated hardening PR than special-case just the codegen job.

.github/workflows/test.yml
```diff
@@ -132,3 +132,28 @@ jobs:
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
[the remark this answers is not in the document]
immrsd: Same as the persist-credentials thread — every action across our workflows uses mutable major tags today, and we don't run Dependabot to bump SHA pins, so pinning just this job would be inconsistent and the pins would go stale. Better handled repo-wide on its own.

scripts/gaussian_codegen/cdf/emit_test_vectors.py
```diff
@@ -0,0 +1,269 @@
...
+(10^9) using the mpmath oracle. Saturation cases (|z| > 6.3) get exact
+endpoint values so the test exercises that branch in cdf.
+
+Two output files are produced from the same source-of-truth case list:
+- `tests/sd29x9_tests/cdf_test_vectors.move` — all cases (positive + negative).
+- `tests/ud30x9_tests/cdf_test_vectors.move` — positive subset.
+
+Both files are `#[test_only]`; each has one `#[test]` driver that iterates the
+table and asserts `|actual − expected| ≤ TOLERANCE` (5 ULP).
+"""
```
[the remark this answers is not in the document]
immrsd: We don't run ruff in CI (no ruff config either), so nothing's actually failing here. The ×/≤/− are deliberate math notation and they also land in the generated .move docs, so I'd rather keep them readable than ASCII-fy. Leaving as-is.

.github/workflows/test.yml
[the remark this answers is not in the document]
immrsd: Yes please

scripts/gaussian_codegen/cdf/validate.py
```diff
@@ -0,0 +1,215 @@
...
+WAD = constants.WAD
+SCALE = constants.SCALE_DECIMAL
+MAX_Z_RAW = constants.MAX_Z_RAW  # 6.3 at 10^9
+MAX_Z_FLOAT = float(constants.MAX_Z)
+HALF_SCALE = SCALE // 2  # Φ(0) bit-exact
+
+COEFF_PATH = (
+    REPO_ROOT / "math" / "fixed_point" / "sources" / "internal" / "cdf_coefficients.move"
+)
+TARGET_ERROR_ULP = 5  # ≤ 5 ULP at 10^-9 (== INV-17 contract)
```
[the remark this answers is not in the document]
immrsd: Done — validate.py now parses MAX_Z_RAW out of the committed cdf_coefficients.move and uses it for both the saturation cutoff and the sweep grid, so it tracks the on-chain bound instead of the Python copy.

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,196 @@
...
+/// product `a × b` fits in `u256`.
+///
+/// #### Aborts
+/// - Aborts if `d == 0`.
+public(package) fun mul_div_nearest_u256(a: u256, b: u256, d: u256): u256 {
+    let prod = a * b;
+    let quot = prod / d;
+    let rem = prod - quot * d;
+    // Half-up: round when remainder ≥ d - remainder, i.e. 2 × rem ≥ d.
+    if (rem * 2 >= d) quot + 1 else quot
```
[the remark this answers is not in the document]
immrsd: Done — switched to rem >= d - rem (equivalent, no rem*2 overflow). Also added direct tests for the helper: below/at/above-half, exact, d==0 abort, and a near-u256::max regression the old form would've aborted on.

scripts/gaussian_codegen/cdf/derive.py
```diff
@@ -0,0 +1,242 @@
...
+    wad = mpf(constants.WAD)
+    max_mag = mpf(0)
+    for c in num + den:
+        m = abs(c) * wad
+        if m > max_mag:
+            max_mag = m
+    print(f"Max |coeff| × WAD = {float(max_mag):.3e} (must be < 2^128 ≈ 3.4e38)")
+    if max_mag >= mpf(2) ** 128:
+        print("FAIL: at least one coefficient does not fit u128 at WAD scale", file=sys.stderr)
+        return 1
```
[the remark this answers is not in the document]
immrsd: Handled in validate.py rather than derive.py — its integer mirror already runs the full Horner over the grid, so it now fails if any multiply/add/final-scale exceeds 2^256. Avoided duplicating the simulation in derive.py.

math/fixed_point/codegen/cdf/emit_test_vectors.py
```diff
@@ -0,0 +1,269 @@
...
+
+Picks a deterministic, hand-curated set of test inputs:
+- Well-known critical points (0, ±0.25, ±0.5, ±1, ±2, ±3, ±4, ±5, ±6, ±6.299,
+  ±6.3, ±6.301, ±7) — exercises symmetry, the saturation boundary on both
+  sides, and the Φ(0) bit-exact case.
+- 16 evenly spaced points across [0, max_z] for breadth.
+
+For each (z, sign) the expected Φ(z) value is computed at the UD30x9 raw scale
+(10^9) using the mpmath oracle. Saturation cases (|z| > 6.3) get exact
+endpoint values so the test exercises that branch in cdf.
```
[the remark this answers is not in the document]
immrsd: Fixed — reworded to |z| >= 6.3 (the inclusive quantized boundary).

math/fixed_point/codegen/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+Three stages, four scripts. Generation flows left to right; `validate.py` is an
+independent re-check that closes the loop.
+
+```
+derive.py ──► .derive_output.json ──┬─► emit_coefficients.py ──► cdf_coefficients.move
+(AAA fit)        (chosen coeffs)     └─► emit_test_vectors.py  ──► cdf_test_vectors.move (×2)
+
+validate.py ──► parses the committed cdf_coefficients.move, re-runs it, checks vs scipy
+```
```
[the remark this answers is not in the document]
immrsd: Done — tagged both fenced blocks as text.

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,199 @@
...
+/// Compute `(a × b) / d` rounded to the nearest integer, half-up (ties away
+/// from zero).
+///
+/// Used to cast a WAD-scale rational `N(z) / D(z)` to a `10^9`-scale value in a
+/// single rounding step. Caller guarantees `d > 0` and that the full-width
+/// product `a × b` fits in `u256`.
+///
+/// #### Aborts
+/// - Aborts if `d == 0`.
+public(package) fun mul_div_nearest_u256(a: u256, b: u256, d: u256): u256 {
```
0xNeshi: In the PR description:
> mul_div_nearest_u256 (L117): round(a·b/d) half-up; inlined to avoid a
math/core dependency (D6).

What's the problem with the math/core dependency? `math/fixed_point` already has it

math/fixed_point/sources/sd29x9/sd29x9_base.move
```diff
@@ -132,6 +140,51 @@ public fun ceil(x: SD29x9): SD29x9 {
...
+/// - `|cdf(z) + cdf(z.negate()) - 1| ≤ 5 ULP` for non-saturated inputs.
+///
+/// #### Aborts
+/// - Does not abort for any `SD29x9` input under the committed, validated
+///   coefficients. The implementation carries internal integrity asserts
+///   (`EInternalNegSubUnderflow` here, plus `cdf::EInternalNumNegative` /
+///   `cdf::EInternalDenNonPositive` in the evaluator) as defense-in-depth
+///   against a corrupted regenerated coefficient table; these cannot fire for
+///   the shipped coefficients.
+public fun cdf(z: SD29x9): SD29x9 {
```
0xNeshi, line 193: Sanity check: did you run `/sui-btt` and `/sui-basic-review` skills?

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,199 @@
...
+    SignedScaled256 { mag, neg: false }
+}
+
+/// Wrap a `(u128 magnitude, bool sign)` coefficient encoding, promoting the
+/// magnitude to `u256` and canonicalizing zero to `(0, false)`.
+public(package) fun signed_from_coeff(mag: u128, neg: bool): SignedScaled256 {
+    let mag_u256 = mag as u256;
+    SignedScaled256 {
+        mag: mag_u256,
+        neg: if (mag_u256 == 0) false else neg,
```
0xNeshi:
```suggestion
        neg: mag_u256 != 0 && neg,
```

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,199 @@
...
+
+public(package) fun is_neg(x: &SignedScaled256): bool { x.neg }
+
+// === Arithmetic ===
+
+/// Sign-magnitude addition.
+/// - Same-sign: add magnitudes, keep the sign.
+/// - Opposite-sign: subtract the smaller from the larger, inherit the larger's sign.
+/// - Exact cancellation: return canonical zero.
+public(package) fun signed_add(a: SignedScaled256, b: SignedScaled256): SignedScaled256 {
```
0xNeshi: nit: can't the function be named just `add`? It's obvious from param types that it produces `SignedScaled256`:
```rust
    let s1 = SignedScaled256 { mag: 1, neg: false };
    let s2 = SignedScaled256 { mag: 1, neg: false };
    let s = s1.add(s2); // result type is obvious
```

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,199 @@
...
+
+/// Polynomial must have at least one coefficient
+#[error(code = 0)]
+const EEmptyPolynomial: vector<u8> = "Polynomial must have at least one coefficient";
+
+// === Constants ===
+
+const WAD_U256: u256 = 1_000_000_000_000_000_000;
+
+// === Sign-Magnitude Value Type ===
```
0xNeshi:
```suggestion
// === Structs ===
```
following `/move-quality` convention

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,199 @@
...
+/// The caller must keep magnitudes bounded so that the full-width product
+/// `a.mag × b.mag` fits in `u256` (`< 2^256`); this is **not** checked here, for
+/// efficiency. For the CDF on `|z| ≤ 6.3` the peak intermediate is ~`1.4 × 10^38`,
+/// ~39 orders of magnitude below `u256::max ≈ 1.16 × 10^77`. A new consumer
+/// (e.g. `pdf`, `inverse_cdf`) with different coefficients or a wider input
+/// domain must re-establish this bound before reusing the evaluator.
+public(package) fun signed_mul_wad(a: SignedScaled256, b: SignedScaled256): SignedScaled256 {
+    let mag = (a.mag * b.mag) / WAD_U256;
+    let neg = if (mag == 0) false
+    else (a.neg != b.neg);
```
0xNeshi:
```suggestion
    let neg = mag != 0 && a.neg != b.neg;
```

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,199 @@
...
+/// #### Aborts
+/// - Aborts with `EEmptyPolynomial` if `$len == 0`.
+public(package) macro fun horner_eval(
+    $z: SignedScaled256,
+    $len: u64,
+    $coeff_at: |u64| -> (u128, bool),
+): SignedScaled256 {
+    let z = $z;
+    let len = $len;
+    openzeppelin_fp_math::horner::assert_polynomial_nonempty(len);
```
0xNeshi: nit: can't a local macros be invoked without the fully-qualified path, in this case with just `assert_polynomial_nonempty(len);`?

math/fixed_point/sources/sd29x9/sd29x9_base.move
```diff
@@ -31,6 +32,13 @@ const ENegativeSqrt: vector<u8> = "Cannot compute square root of a negative valu
 #[error(code = 4)]
 const ELogUndefined: vector<u8> = "Logarithm is undefined: input must be strictly positive";
 
+/// `cdf_nonneg_raw` returned a value below `Φ(0) = 0.5`, which would make the
+/// negative-input sign-flip subtraction `10^9 - phi` produce a result greater
+/// than `0.5`. Defense-in-depth against an AAA-fit regression.
+#[error(code = 5)]
+const EInternalNegSubUnderflow: vector<u8> =
+    "CDF sign-flip subtraction underflowed: cdf_nonneg_raw returned below 0.5";
```
0xNeshi: The chance that this error appears is almost 0%. However, if it does, the error message would contain a mention of an internal function, which is something the end user shouldn't care about.

As a matter of principle, consider rewording the error message to keep clarity, but remove mention of the `cdf_nonneg_raw` internal function.

math/fixed_point/sources/sd29x9/sd29x9_base.move
```diff
@@ -132,6 +140,51 @@ public fun ceil(x: SD29x9): SD29x9 {
...
+/// - `z`: Input value.
+///
+/// #### Returns
+/// - The probability `Φ(z)` as a non-negative `SD29x9` in `[0, 1]`.
+///
+/// #### Behavior
+/// - Saturates exactly to `0` for `z ≤ -6.3` and to `1` for `z ≥ 6.3`. At
+///   those bounds `Φ` is already within `~10⁻¹⁰` of the saturated value,
+///   well below the output's `10⁻⁹` resolution.
+/// - `Φ(0)` is bit-exactly `0.5`.
```
0xNeshi:
```suggestion
/// - `Φ(0)` is exactly `0.5`.
```
"bit-exactly" sounds confusing, unless it's a standard term for this kind of math, consider removing for clarity

math/fixed_point/sources/sd29x9/sd29x9_base.move
```diff
@@ -205,6 +258,18 @@ public fun is_zero(x: SD29x9): bool {
...
 }
 
+/// Checks whether a value is strictly less than zero (sign bit set).
+///
+/// #### Parameters
+/// - `x`: Input value.
+///
+/// #### Returns
+/// - `true` if `x < 0`, otherwise `false`. Returns `false` for zero, positive
+///   values, and `max()`; returns `true` for any negative value including `min()`.
```
0xNeshi:
```suggestion
/// - `true` if `x < 0`, otherwise `false`.
```
clear enough

0xNeshi:
Btw, excellent work! :rocket:

math/fixed_point/codegen/cdf/validate.py
```diff
@@ -0,0 +1,241 @@
...
+    if prod > U256_MAX:
+        raise RuntimeError(f"u256 overflow in signed_mul_wad: {am} * {bm}")
+    mag = prod // WAD
+    return _canonicalize((mag, an ^ bn))
+
+
+def horner_eval(z: SignedInt, coeffs: list[SignedInt]) -> SignedInt:
+    if not coeffs:
+        raise RuntimeError("empty polynomial")
+    acc = coeffs[-1]
```
0xNeshi:
```suggestion
    acc = _canonicalize(coeffs[-1])
```
probably not a problem, but for consistency with the on-chain code, the input should be canonicalized:

math/fixed_point/sources/internal/horner.move
immrsd: Fair point — the "avoid a math/core dependency" rationale was stale, the dependency already exists. Replaced the local helper with `u256::mul_div(n, 10^9, d, rounding::nearest())` and removed `mul_div_nearest_u256` plus its direct tests. Tie semantics are identical (Nearest breaks ties up), and all committed CDF test vectors pass bit-identical. The Python mirror in validate.py is unchanged logic-wise, just re-pointed at the math/core function.

math/fixed_point/sources/internal/horner.move
immrsd: Done

math/fixed_point/sources/internal/horner.move
immrsd: Done — and renamed the whole family for consistency (`signed_zero`/`signed_from_unsigned`/`signed_from_coeff`/`signed_add_coeff`/`signed_mul_wad` → `zero`/`from_unsigned`/`from_coeff`/`add_coeff`/`mul_wad`), including the Python mirror in validate.py.

math/fixed_point/sources/internal/horner.move
immrsd: Done

math/fixed_point/sources/internal/horner.move
immrsd: Done

math/fixed_point/sources/internal/horner.move
immrsd: It can — names in macro bodies resolve at the definition site (math/core's `mul_div!` macro already calls `mul_div_inner` unqualified across modules). Dropped the fully-qualified paths from all four calls in the macro body.

math/fixed_point/sources/sd29x9/sd29x9_base.move
immrsd: Done — reworded to drop the internal function name.

math/fixed_point/sources/sd29x9/sd29x9_base.move
immrsd: Done — switched to "exactly" here and in the other cdf doc occurrences (ud30x9_base.move, cdf.move).

math/fixed_point/sources/sd29x9/sd29x9_base.move
immrsd: Done

math/fixed_point/sources/sd29x9/sd29x9_base.move
immrsd: Ran both, pushed missing test cases

math/fixed_point/codegen/cdf/validate.py
immrsd: Done — and added a mirror test for it (`(0, True)` leading coefficient → canonical `(0, False)`), matching the existing Move-side `horner_eval_zero_polynomial_canonicalizes` test.

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+// === Accessors ===
+
+public(package) fun mag(x: &SignedScaled256): u256 { x.mag }
+
+public(package) fun is_neg(x: &SignedScaled256): bool { x.neg }
+
+// === Arithmetic ===
```
0xNeshi, line 77: Sanity check: run `/move-quality` skill

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,176 @@
...
+
+// === Errors ===
+
+/// Polynomial must have at least one coefficient
+#[error(code = 0)]
+const EEmptyPolynomial: vector<u8> = "Polynomial must have at least one coefficient";
+
+// === Constants ===
+
+const WAD_U256: u256 = 1_000_000_000_000_000_000;
```
0xNeshi: Add short explanation what "WAD" means and about its purpose

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,176 @@
...
+/// dependency: it provides only the reusable numeric primitives, so each
+/// consumer supplies its own coefficient table. Concretely it offers a
+/// `SignedScaled256` value, sign-aware add/multiply, and the `horner_eval!`
+/// macro that evaluates a polynomial via Horner's method given a `(u128, bool)`
+/// coefficient accessor.
+module openzeppelin_fp_math::horner;
+
+// === Errors ===
+
+/// Polynomial must have at least one coefficient
```
0xNeshi: nit: add missing dots to end of comments

ericnordelo (commented):
The implementation itself looks good, great job! but I think we should reconsider how we include and present it in the library. Right now the PR includes the full offline coefficient-generation pipeline alongside the runtime Move implementation. That creates a lot of noise in this repository. Most library users do not need to understand or maintain the AAA fitting process, they care about the public `cdf` API, its behavior, its error bounds, its saturation semantics, and whether the implementation is deterministic and safe to use on-chain.

I think we should separate the runtime library surface from the coefficient-generation tooling. The generated coefficients and the Move evaluator belong here, but the heavier research/codegen pipeline could live outside this package, possibly in a dedicated OpenZeppelin repository or another clearly separated tooling location. If we keep any generation or validation code here, it should be minimal and justified by reproducibility or CI needs.

We should also put more emphasis on documenting the useful properties of the CDF implementation:


- maximum absolute error / ULP bound
- valid input and output ranges
- saturation behavior at `|z| >= 6.3`
- exact `Φ(0) = 0.5`
- symmetry for signed inputs
- deterministic, pure, object-free execution
- intended use cases and limitations


Those details should be easy to find in the README, public doc comments, examples, and eventually the docs site, because this is what will make users trust and integrate the function.

Finally, I think comments should be written (mostly) for users, and agents trying to understand the flow. Generated markers like `AUTO-GENERATED — do not hand-edit` may be useful, but the source should avoid over-explaining the offline pipeline in places where users are only trying to understand the runtime behavior. In this case for example users can’t modify that code, and we maintainers ofc should not modify these values manually.

math/fixed_point/sources/internal/horner.move
immrsd: Done

math/fixed_point/sources/internal/horner.move
immrsd: Added a doc comment with explanation why it's needed here

math/fixed_point/sources/internal/horner.move
immrsd: Ran the skill and fixed issues

math/fixed_point/README.md
```diff
@@ -120,6 +121,44 @@ let half = sd29x9::wrap(500_000_000, false); // 0.5
...
+- **Saturation**: exactly `0` for `z ≤ -6.3` (SD29x9) and exactly `1` for
+  `z ≥ 6.3`.
+- **Φ(0)**: exactly `0.5`.
+- **Symmetry**: `cdf(z) + cdf(z.negate())` is exactly `1` for every `SD29x9`
+  input.
+- **Execution**: pure, deterministic, and object-free integer math - no storage,
+  no Sui objects; identical inputs always yield identical outputs.
+
+```move
+use openzeppelin_fp_math::{sd29x9, ud30x9};
```
0xNeshi:
```suggestion
use openzeppelin_fp_math::sd29x9;
use openzeppelin_fp_math::ud30x9;
```
nit: Move formatter would output this

math/fixed_point/README.md
```diff
@@ -138,6 +177,18 @@ let third_down = one.div_trunc(ud30x9::wrap(3000000000)); // 0.333333333
...
+
+The standard-normal CDF (`cdf`) is backed by an AAA-rational approximation whose
+coefficients and test vectors are generated offline and must **not** be
+hand-edited (each carries an `AUTO-GENERATED` banner):
+
+- `sources/internal/cdf_coefficients.move`
+- `tests/{sd29x9_tests,ud30x9_tests}/cdf_test_vectors.move`
+
+To regenerate them — or to re-validate the committed coefficients against
+`scipy` — see [`scripts/gaussian_codegen/`](../../scripts/gaussian_codegen/README.md).
```
0xNeshi:
```suggestion
To regenerate them - or to re-validate the committed coefficients against
`scipy` - see [`scripts/gaussian_codegen/`](../../scripts/gaussian_codegen/README.md).
```
double check no em dashes are used anywhere else

STYLEGUIDE.md
```diff
@@ -203,6 +203,10 @@ Examples of required rewrites:
   `#[test_only]` helpers (compiled only in test mode) and `#[test]` functions
   that exercise private items unreachable from `tests/`. Place all `#[test_only]`
   and `#[test]` items under the `// === Test-Only Helpers ===` section (item 13).
+- Do **not** put `#[test_only]` on the module declaration of a file under `tests/` -
+  those files are compiled only in test mode by location, so a module-level
+  `#[test_only]` is redundant. The attribute belongs only on test-only items inside a
+  *production* module.
```
0xNeshi, line 209: 👍

0xNeshi (approved):
Nits, approving in advance.
Good work :+1:

bidzyyys:
@coderabbitai review again

math/fixed_point/README.md
immrsd: Fixed

math/fixed_point/README.md
immrsd: Fixed. Double-checked the rest while I was at it - there are no em dashes left in any `.move`. A handful remain in the `scripts/gaussian_codegen` docs (Python/Markdown tooling, not Move); happy to sweep those too if we want them gone.

math/fixed_point/sources/internal/cdf.move
```diff
@@ -0,0 +1,129 @@
...
+
+// === Accessors ===
+
+/// `Φ(0)` at the raw scale - the lower bound of `cdf_nonneg_raw`'s return
+/// range. Exposed so callers can check the `phi ≥ 0.5` contract against the
+/// same constant the evaluator uses.
+public(package) fun half_raw(): u128 { HALF_RAW }
+
+// === CDF Central-Domain Helper ===
+
```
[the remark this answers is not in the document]
immrsd: `Accessors` and `CDF Central-Domain Helper` are feature sub-groups inside `// === Package Functions ===`, not top-level sections - the STYLEGUIDE section-ordering rule explicitly allows these and says not to remove them. The top-level order is already canonical, and `// === Imports ===` isn't a required delimiter (imports are prettier-grouped). Leaving as-is.

math/fixed_point/sources/internal/horner.move
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+// === Accessors ===
+
+public(package) fun mag(x: &SignedScaled256): u256 { x.mag }
+
+public(package) fun is_neg(x: &SignedScaled256): bool { x.neg }
+
+// === Arithmetic ===
```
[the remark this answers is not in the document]
immrsd: `Constructors` / `Accessors` / `Arithmetic` are sub-groups within `// === Package Functions ===`, which STYLEGUIDE explicitly permits (it even says to keep them). The top-level ordering is already canonical, so no change here.

math/fixed_point/tests/gaussian_tests/horner_tests.move
```diff
@@ -0,0 +1,281 @@
...
+fun pos(mag: u256): SignedScaled256 { from_unsigned(mag) }
+
+// All test magnitudes fit u128, so signed values can be built directly via
+// `from_coeff`; zero is routed through `zero()` to match the canonical
+// encoding.
+fun signed(mag: u256, neg: bool): SignedScaled256 {
+    if (mag == 0) zero() else from_coeff(mag as u128, neg)
+}
+
+// === zero / from_* / accessors ===
```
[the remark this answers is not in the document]
immrsd: These group the tests by the function under test - the same convention the existing `math/core` suites use (`common_tests.move`, `math_ops.move`). STYLEGUIDE allows sub-group headers, so keeping them.

math/fixed_point/tests/sd29x9_tests/cdf_test_vectors.move
```diff
@@ -0,0 +1,89 @@
+// AUTO-GENERATED - do not hand-edit.
+// Source: scripts/gaussian_codegen/cdf/emit_test_vectors.py (oracle: mpmath ncdf at 100 dps)
+
+/// Deterministic test vectors for `sd29x9_base::cdf`. Each row asserts the
+/// result of `sd29x9::wrap(z_raw, neg).cdf()` matches `expected` to within
+/// `TOLERANCE` raw SD29x9 ULPs (== 5 × 10^-9 absolute).
+#[test_only]
+module openzeppelin_fp_math::sd29x9_cdf_test_vectors;
+
```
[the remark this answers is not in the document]
immrsd: The module-level `#[test_only]` is already gone (33b2a02). This file is generated and its layout is minimal and already ordered, so I'm not restructuring it. No change.

math/fixed_point/tests/sd29x9_tests/cdf_test_vectors.move
```diff
@@ -0,0 +1,89 @@
...
+use openzeppelin_fp_math::sd29x9;
+
+const TOLERANCE: u128 = 5; // ≤ 5 ULP at UD30x9 scale (10^-9)
```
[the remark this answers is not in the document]
immrsd: Good catch - corrected the label to `SD29x9 scale` in the emitter and regenerated.

math/fixed_point/tests/sd29x9_tests/cdf_tests.move
```diff
@@ -0,0 +1,302 @@
...
+// 5 ULP at the SD29x9 scale (≡ 5 × 10^-9 absolute), per the accuracy contract.
+const TOLERANCE: u128 = 5;
+
+// Reference Φ values at the UD30x9 raw scale (rounded from scipy/mpmath at
+// 100 dps). The well-known-value tests read like statistics-table entries.
+const PHI_1_RAW: u128 = 841_344_746;
+const PHI_2_RAW: u128 = 977_249_868;
+const PHI_3_RAW: u128 = 998_650_102;
+
+// === Bit-exact / well-known points ===
```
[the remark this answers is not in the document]
immrsd: Same as the other test modules - grouped by scenario, matching the repo's existing test style. STYLEGUIDE permits these sub-group headers, so leaving as-is.

math/fixed_point/tests/ud30x9_tests/cdf_test_vectors.move
```diff
@@ -0,0 +1,61 @@
...
+const ETestCaseFailed: vector<u8> =
+    "cdf test vector mismatch: |actual - expected| exceeded TOLERANCE";
+
+public struct TestCase has copy, drop {
+    z_raw: u128,
+    expected: u128,
+}
+
+#[test]
+fun cdf_vectors_match_oracle() {
```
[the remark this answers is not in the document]
immrsd: This is generated output - the layout (banner, module doc, `use`, constants, struct, one `#[test]`) is already minimal and ordered. Bolting the full section scaffold onto a generated file isn't worth it and STYLEGUIDE doesn't require it here. No change.

math/fixed_point/tests/ud30x9_tests/cdf_test_vectors.move
```diff
@@ -0,0 +1,61 @@
+// AUTO-GENERATED - do not hand-edit.
+// Source: scripts/gaussian_codegen/cdf/emit_test_vectors.py (oracle: mpmath ncdf at 100 dps)
+
+/// Deterministic test vectors for `ud30x9_base::cdf`. Each row asserts the
+/// result of `ud30x9::wrap(z_raw).cdf()` matches `expected` to within
+/// `TOLERANCE` raw UD30x9 ULPs (== 5 × 10^-9 absolute).
+#[test_only]
+module openzeppelin_fp_math::ud30x9_cdf_test_vectors;
```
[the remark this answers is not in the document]
immrsd: Already removed in 33b2a02.

math/fixed_point/tests/ud30x9_tests/cdf_tests.move
```diff
@@ -0,0 +1,161 @@
...
+const TOLERANCE: u128 = 5;
+
+// Reference Φ values at the UD30x9 raw scale (rounded from scipy/mpmath at
+// 100 dps). The well-known-value tests read like statistics-table entries.
+const PHI_1_RAW: u128 = 841_344_746;
+const PHI_2_RAW: u128 = 977_249_868;
+const PHI_3_RAW: u128 = 998_650_102;
+
+// === Bit-exact / well-known points ===
+
```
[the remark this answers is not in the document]
immrsd: Same as the sd29x9 test module - scenario sub-groups, consistent with the rest of the repo. Keeping as-is.

scripts/gaussian_codegen/cdf/emit_coefficients.py
```diff
@@ -0,0 +1,156 @@
...
+
+/// Denominator magnitudes (ascending power order).
+public(package) fun cdf_den_mags(): vector<u128> {{ DEN_MAGS }}
+
+/// Denominator sign flags (ascending power order); index `i` paired with `cdf_den_mags()[i]`.
+public(package) fun cdf_den_negs(): vector<bool> {{ DEN_NEGS }}
+
+/// Saturation threshold |z| at the raw `10^9` scale (`6_300_000_000`).
+public(package) fun max_z_raw(): u128 {{ MAX_Z_RAW }}
+"""
```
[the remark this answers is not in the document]
immrsd: The generated `cdf_coefficients.move` already has `// === Constants ===` and `// === Package Functions ===` in canonical order - empty `Imports`/`Errors` headers aren't required. And the accessors are trivial getters whose summary already states what they return, so per STYLEGUIDE they don't need `#### Parameters/Returns/Aborts`. Leaving the template as-is.

scripts/gaussian_codegen/cdf/emit_test_vectors.py
```diff
@@ -0,0 +1,264 @@
...
+        {case_lines},
+    ];
+    cases.destroy!(|case| {{
+        let z = sd29x9::wrap(case.z_raw, case.neg);
+        let actual = z.cdf().unwrap();
+        let diff = if (actual >= case.expected) actual - case.expected else case.expected - actual;
+        assert!(diff <= TOLERANCE, ETestCaseFailed);
+    }});
+}}
+"""
```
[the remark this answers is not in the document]
immrsd: The `#[test_only]` on the generated test modules is already gone (33b2a02). Skipping the section-scaffold change for the same reason as the other test files - the generated layout is fine. The `×` is intentional math notation, and we don't run Ruff in CI.

scripts/gaussian_codegen/tests/test_validate_mirror.py
```diff
@@ -0,0 +1,137 @@
...
+def test_cdf_simulate_phi0(coeffs):
+    num, den = coeffs
+    assert v.cdf_simulate(0, False, num, den) == v.HALF_SCALE
+
+
+def test_cdf_simulate_saturation(coeffs):
+    num, den = coeffs
+    assert v.cdf_simulate(v.MAX_Z_RAW, False, num, den) == v.SCALE
+    assert v.cdf_simulate(v.MAX_Z_RAW, True, num, den) == 0
+
```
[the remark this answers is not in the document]
immrsd: The committed `MAX_Z_RAW` is already drift-guarded: `validate.py` parses it out of `cdf_coefficients.move` and runs the full sweep against it in the CI gate. This unit test just exercises `cdf_simulate`'s saturation branch, so re-parsing the committed value here would only duplicate the gate. Leaving it.

math/fixed_point/README.md
0xNeshi: I'm fine either way, it's only important that publicly-facing code/docs use regular dashes

immrsd:
@coderabbitai review once again

math/fixed_point/README.md
```diff
@@ -120,6 +122,45 @@ let half = sd29x9::wrap(500_000_000, false); // 0.5
...
+
+- **Accuracy**: max absolute error `≤ 5 × 10⁻⁹` (5 ULP at the `10⁹` scale);
+  empirical worst case `~7 × 10⁻¹⁰`.
+- **Domain**: effective input range `|z| ≤ 6.3`; beyond that the result
+  saturates.
+- **Saturation**: exactly `0` for `z ≤ -6.3` (SD29x9) and exactly `1` for
+  `z ≥ 6.3`.
+- **Φ(0)**: exactly `0.5`.
+- **Symmetry**: `cdf(z) + cdf(z.negate())` is exactly `1` for every `SD29x9`
+  input.
```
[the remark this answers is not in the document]
immrsd: Good point - `negate()` aborts on `min()` (no representable `+2^127`), so the identity can't be evaluated there. Reworded to exclude `min()`.

math/fixed_point/tests/ud30x9_tests/cdf_tests.move
```diff
@@ -0,0 +1,161 @@
...
+        MAX_Z_RAW - 1,
+        MAX_Z_RAW,
+        MAX_Z_RAW + 1,
+        7 * SCALE,
+    ];
+    probes.destroy!(|raw| {
+        let via_unsigned = fixed(raw).cdf().unwrap();
+        let via_signed = pos(raw).cdf().unwrap();
+        assert_eq!(via_unsigned, via_signed);
+    });
```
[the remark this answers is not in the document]
immrsd: Agreed - the comment claimed `into_SD29x9` but the code used the `pos` helper. Switched the signed side to `fixed(raw).into_SD29x9().cdf()` so it actually exercises the conversion, and dropped the now-unused `pos` import.

scripts/gaussian_codegen/cdf/emit_coefficients.py
```diff
@@ -0,0 +1,162 @@
...
+
+def main(argv: Sequence[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description="Emit cdf_coefficients.move.")
+    parser.add_argument("--input", type=pathlib.Path, default=DERIVE_OUTPUT_PATH)
+    parser.add_argument("--output", type=pathlib.Path, default=COEFF_OUTPUT_PATH)
+    parser.add_argument(
+        "--check",
+        action="store_true",
+        help="Verify the committed file matches freshly generated output; exit 1 on "
+        "drift, do not write. Requires a prior `derive` run for the JSON input.",
```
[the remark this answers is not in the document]
immrsd: Fixed - the committed `.derive_output.json` is enough for `--check`, so reworded the help to say the input JSON just needs to exist.

scripts/gaussian_codegen/cdf/validate.py
```diff
@@ -0,0 +1,259 @@
...
+
+    text = args.coeffs.read_text(encoding="utf-8")
+    num, den = parse_coefficients(text)
+    max_z_raw = _parse_u128_const(text, "MAX_Z_RAW")
+    print(
+        f"Parsed {len(num)} numerator + {len(den)} denominator coefficients "
+        f"(MAX_Z_RAW = {max_z_raw}) from {args.coeffs.relative_to(REPO_ROOT)}"
+    )
+
+    grid = np.linspace(0.0, max_z_raw / SCALE, args.n)
```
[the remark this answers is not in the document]
immrsd: Done - added an `args.n >= 2` guard that fails fast, so `--n 0` (empty sweep, false PASS) and `--n 1` no longer slip through.

scripts/gaussian_codegen/shared/move_emit.py
```diff
@@ -0,0 +1,131 @@
...
+            f"prettier not found at {prettier}; run `npm ci` at the repo root "
+            "(installs @mysten/prettier-plugin-move, the Move formatter)"
+        )
+    result = subprocess.run(
+        [str(prettier), "--stdin-filepath", str(path)],
+        input=text,
+        capture_output=True,
+        text=True,
+        cwd=repo_root,
+    )
```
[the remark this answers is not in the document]
immrsd: Done - added a 30s timeout and routed `TimeoutExpired` through the same `RuntimeError` path. Left the Ruff S603 note alone: the prettier path is our own, not untrusted input, and we don't run Ruff in CI.

ericnordelo (approved):
Great job @immrsd . LGTM!

PR #360, opened 2026-06-09
[N-03] Math Core Audit

ericnordelo:
Fixes [N-03](https://audits.openzeppelin.com/sui-contracts/sui-01-09-core-math-audit/issue/style-and-convention-inconsistencies-across-the-math-package-9b37e61d)

## Summary by CodeRabbit

* **Chores**
  * Updated Sui CLI version to 1.72.5.

* **Documentation**
  * Enhanced inline documentation and comments for mathematical operations and vector sorting functionality.
  * Clarified bit-width constants across numeric types.
  * Improved setup instructions to reflect the updated CLI version requirement.

PR #363, opened 2026-06-10
[N-04] Core Math Audit

ericnordelo:
## Rationale: inlined `median!` macro + precompiled wrapper functions

**Approach.** We benchmarked eight median implementations with `sui move test -s csv`
(deterministic gas; inputs spanning 100–5000 elements; ascending, reverse-sorted, and
duplicate-heavy patterns), measuring both execution gas and compiled bytecode per call site.

### What the benchmarks showed

- **Pivot caching dominates:** reading `vec[pivot_index]` on every comparison was the single
  largest cost; caching it in a local cut the `u64` selection ~79%. An index-vector quickselect
  variant only looked competitive because it incidentally did this — compared fairly, plain
  value-based quickselect wins at every width.
- **Widening is the macro's bottleneck:** a generic `median!` that converts to `u256` and
  delegates to one function pays the widening on every narrow-width call. Move macros cannot
  dispatch on `$Int` to width-specific functions (type errors at expansion), so the only way to
  run selection at native width through one generic entry is to inline the whole algorithm in
  the macro and let expansion monomorphize it.
- **Cloning via dereference:** `*vec` copies the whole vector in one instruction (~1 gas/KB) vs
  `map_ref!`'s per-element loop — worth ~20% end-to-end on its own.

### Constraints that shaped the design

A public macro expands in the **caller's** module, so it may only reference public symbols — it
cannot use private helper macros or the module's private error constant. The empty-input abort
is therefore routed through `median_u8`, which validates before expanding the macro, preserving
the canonical `EMedianOfEmptyVector` everywhere. Inlining also removes the need for any engine
function, so every public entry takes an immutable reference — nothing is consumed or mutated.

### The trade-off, and why wrappers exist

Each `median!` call site adds ~750–870 B of compiled bytecode (comparable to the existing
`quick_sort!` at ~750 B), which counts against Sui's package-size cap. The
`median_u8`…`median_u256` wrappers precompile the same expansion once in the library, costing
call sites only a regular function call — measured at identical gas to within a few units.
Users pick per call site; the docs explain the trade-off.

### Results

vs the previous implementation (n=1001, normalized gas):

| Path | Before | After | Δ |
| ---- | ------ | ----- | - |
| `median!` u64 | 91.5k | **6.9k** | −92% |
| `median!` u128 | 95.5k | **7.9k** | −92% |
| `median!` u256 | 103.5k | **46.3k** | −55% |
| `median!` u64, sorted input (n=5001) | 120.6k | **8.0k** | −93% |

Validated by 775 tests including randomized cross-checks against a sorted reference, oracle-free
property tests, and an external consumer package proving cross-package macro expansion and
abort-code behavior.

One notable issue surfaced and fixed along the way: with the algorithm inlined, ~120 macro call
sites in our own test suite exceeded the Move VM's 10 MB package arena
(`PACKAGE_ARENA_LIMIT_REACHED`) — behavioral coverage now runs through the wrappers, with a
compact per-width macro section (the same guidance downstream test suites should follow).

## Summary by CodeRabbit

* **New Features**
  * Introduced precompiled vector median wrapper functions for types u8, u16, u32, u64, u128, and u256 with optimized library-level bytecode compilation.
  * Updated `median_u256` to accept non-mutating references for improved efficiency.

* **Documentation**
  * Enhanced vector documentation with median wrapper function details and performance trade-off information.


math/core/sources/vector.move
```diff
@@ -241,160 +290,207 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
-}
-
-// === Concrete Public Functions ===
+    let rounding_mode = $rounding_mode;
+
+    // An empty vector has no median. The check delegates to `median_u8`, whose own emptiness
+    // assertion raises the module's canonical `EMedianOfEmptyVector`: a public macro expands in
+    // the caller's module, where this module's private error constant is not visible, so the
+    // abort must come from a public function. `median_u8` validates before expanding this
+    // macro, which is what guarantees the call below always aborts instead of recursing.
```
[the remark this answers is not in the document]
0xNeshi: Seems like a valid concern: https://github.com/OpenZeppelin/contracts-sui/pull/363/changes#r3387745050

Alternative solution is to wrap `EMedianOfEmptyVector` in a public `assert_not_empty`. It does expose this to outside users, but it isn't a big deal.

math/core/sources/vector.move
```diff
@@ -241,160 +290,207 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
...
+/// Compute the median of a `u128` vector with configurable rounding.
+///
+/// Precompiled wrapper around `median!`: same algorithm, same results, same abort behavior,
+/// but the selection bytecode lives in this module instead of being inlined at the call site.
+/// The input vector is not mutated. See `median!` for the full contract and the bytecode
+/// vs. gas trade-off.
+///
+/// #### Aborts
+/// - `EMedianOfEmptyVector` if `vec` is empty.
+public fun median_u128(vec: &vector<u128>, rounding_mode: RoundingMode): u128 {
```
0xNeshi, line 515: What's the use case for each `median_X` variant, if devs can use the `median!` macro directly?

math/core/sources/vector.move
```diff
@@ -241,160 +290,207 @@ public macro fun quick_sort_by<$T>($vec: &mut vector<$T>, $le: |&$T, &$T| -> boo
 /// ```
 public macro fun median<$Int>($vec: &vector<$Int>, $rounding_mode: RoundingMode): $Int {
```
0xNeshi, line 291: Do we also want `median_by<$Int>`, just like for [quick_sort_by](https://github.com/OpenZeppelin/contracts-sui/blob/a199c9183126ea3baf7c822f7435b027d093224b/math/core/sources/vector.move#L112)?

0xNeshi: Aligns with [claim](https://github.com/OpenZeppelin/contracts-sui/pull/360/changes#diff-dc68b9b3d864cc148d0417baf7f7fac546a4f0bcc300fb2bc6102e898532b8edR24):
> ## Vector operations
> 
> Generic over `u8`..`u256`, with comparator-based sorting available for other types:

math/core/sources/vector.move
ericnordelo: TLDR: Check the Bytecode vs. gas trade-off sections in comments.

Explanation:

is the PR's core trade-off. Gas is identical to within a few units (measured: 6,863 vs 6,864 normalized, u64 n=1001), but each median! call site inlines ~750–870 B of compiled bytecode against Sui's 100 KiB package-size cap (and one VM arena slot in test packages — our own suite hit PACKAGE_ARENA_LIMIT_REACHED before routing coverage through wrappers). A wrapper call costs ~23 B. So: wrappers are the default; the macro is for gas-critical or few-call-site code. The wrapper docs and the #### Bytecode vs. gas trade-off section already say this

math/core/sources/vector.move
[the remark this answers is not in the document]
ericnordelo: Don't you think is enough with updating the wording?

math/core/sources/vector.move
ericnordelo: Sorting needs only an ordering but median additionally needs arithmetic to average, so it is not enough with a simple callback.

math/core/sources/vector.move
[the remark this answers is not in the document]
0xNeshi: It can be fine :+1:

math/core/sources/vector.move
0xNeshi: Right, ignore this then

0xNeshi (approved):

math/core/tests/macros/quick_sort.move
```diff
@@ -823,3 +823,43 @@ fun quick_sort_by_median_of_three_crafted_inputs() {
...
+    // patterns are the coverage target here, and large inputs are exercised by the fixed tests.
+    while (v.length() > 200) {
+        v.pop_back();
+    };
+
+    let mut sorted = v;
+    vector::quick_sort!(&mut sorted);
+
+    let len = sorted.length();
+    assert_eq!(len, v.length());
```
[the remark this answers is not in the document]
ericnordelo: Why do we want to original input available on a tests of random values?

bidzyyys (approved):
LGTM

PR #389, opened 2026-06-16
test: remove #[test_only] from *_tests modules

0xNeshi:
As per [docs](https://move-book.com/testing/testing-basics#test-only-code), the `#[test_only]` attribute is used to only include the annotated code in the final bytecode output if the `mode` is set to `test`, i.e. `sui move build --mode test` or `sui move test` (uses `--mode test` by default).
This is used when some module in the `sources/*` contains test-specific code (see [move best practices](https://docs.sui.io/develop/write-move/move-best-practices#body))

However, any module inside the `tests/*` directory is by default _only ever included_ in the final bytecode if `--mode test` is set (see [Move Book > Packages](https://move-book.com/reference/packages#package-layout-and-manifest-syntax)).
Confirmed this claim locally.

Removing the attribute from modules inside `tests/*` results in no bytecode changes in any mode.

## Summary by CodeRabbit

* **Tests**
  * Removed test-only compilation restrictions across many Move test suites and helper utilities, making these modules available outside test-only build contexts.
  * Updated select test helper constructs to remove test-only gating (including module/fixture annotations), without changing any test logic or assertions.

### DIFF
--- a/contracts/access/tests/access_control_tests.move
+++ b/contracts/access/tests/access_control_tests.move
@@ -38,7 +38,6 @@ public struct RoleY {}
 /// Deploys an `AccessControl` rooted at `ACCESS_CONTROL_TESTS`, shares it,
 /// and advances to a fresh transaction so `take_shared` is immediately
 /// available.
-#[test_only]
 #[allow(lint(share_owned))]
 fun setup(deployer: address, delay: u64): Scenario {
     let mut scenario = test_scenario::begin(deployer);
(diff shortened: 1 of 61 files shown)


bidzyyys (approved):
LGTM

ericnordelo:
Do we have to remove them though? Is there a benefit from removing them that outweights quickly recognizing these items as test only? I think it helps with understanding the code faster visually.

contracts/access/tests/access_control_tests.move
ericnordelo, line 12: We are missing a few like this one

0xNeshi: Actually the attribute is needed here, added comment explaining why

0xNeshi: This might be a Move bug, opened an issue: https://github.com/MystenLabs/sui/issues/26993

ericnordelo (approved):

PR #388, opened 2026-06-16
style: change em dashes (`—`) into regular dashes (`-`)

0xNeshi:
## Summary by CodeRabbit

* **Documentation**
  * Standardized wording and punctuation across repository guides and instructions (STYLEGUIDE.md, ARCHITECTURE.md, AGENTS.md, CONTRIBUTING.md, README.md, llms.txt, and related command/instruction docs) for consistent formatting and clearer guidance.
  * Updated contract/module and example documentation comments and README sections, including minor clarifications around the `/code-quality` workflow.
* **Style**
  * Normalized typographic dash characters and reflowed Markdown/module/test comments throughout the repo (documentation-only).

PR #392, opened 2026-06-16
chore: code-style fixes per STYLEGUIDE

bidzyyys:
General code-style cleanup against `STYLEGUIDE.md`. **Does not close any issue.**

## Changes

- **STYLEGUIDE.md** — codify doc-comments as the API source of truth (rendered by `sui move build --doc`, consumed by AI-integrator tooling):
  - `#### Aborts` lists **every** error a function can raise, including errors propagated from internal calls;
  - state caller preconditions and map each to its error;
  - `#### Parameters` / `#### Returns` only where they add detail beyond the one-line summary (a trivial getter needs neither);
  - **pin-exact exception:** approximation tests that check an on-chain result against an off-chain mathematical reference within a tolerance (e.g. `assert!(diff < epsilon)`) stay bound-based — do not convert to `assert_eq!`.
- **Struct field docs** — document the hot-potato struct fields in `access` (`delayed`/`two_step` `Borrow`, `two_step` `RequestBorrow`) and `fixed_point` `Components`.
- **Test naming** — drop the `test_` prefix from `access_control_tests` functions (descriptive names per STYLEGUIDE Testing), matching the sibling test modules.

## Verification

`sui move test --build-env testnet --lint --warnings-are-errors` and `sui move build --doc --build-env testnet`, all green:
- `contracts/access` — 137 tests
- `math/core` — 753 tests
- `math/fixed_point` — 426 tests

## Out of scope (follow-up)

`access_control_tests` still carries `#[allow(lint(...))]` (4) and `abort 999` sentinels (43); removing them requires restructuring the expected-failure tests (and resolving `share_owned`) and is left to a separate focused PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Expanded public documentation standards to require complete `#### Parameters`, `#### Returns`, and detailed `#### Aborts` listings.
  * Updated API docs across math and access components with explicit abort/error codes and clearer explanations.
  * Added field-level documentation for public structs and refreshed the project’s introductory wording.
* **Tests**
  * Expanded and renamed AccessControl tests, adding more negative and boundary coverage plus a regression for correct pending-state cleanup.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -252,10 +252,11 @@ part of the public API, not decoration. Keep them complete and accurate.
 - Document public functions: include `#### Parameters` and `#### Returns` where
   they add detail beyond the one-line summary - a trivial getter whose summary
   already states what it returns needs neither. Always include an `#### Aborts`
-  section whenever a function can abort, listing **every** error it can raise -
-  including errors propagated from internal calls
-- State caller preconditions explicitly and map each to the error it fails with
-  (e.g. "Aborts with `EUnauthorized` if the caller lacks the role")
+  section whenever a function can abort, listing **every** cause it can raise -
+  including native aborts (e.g. arithmetic overflow, division by zero) and
+  errors propagated from internal calls
+- State caller preconditions explicitly and map each to the error it fails with,
+  one bullet per cause (e.g. "`EUnauthorized` if the caller lacks the role")
 - Keep terminology consistent with the implementation (e.g. avoid documenting
   impossible paths)
 
@@ -270,5 +271,6 @@ part of the public API, not decoration. Keep them complete and accurate.
   /// - Rounded output value.
   ///
   /// #### Aborts
-  /// - Aborts if `value` is zero.
+  /// - `EZeroValue` if `value` is zero.
+  /// - Arithmetic overflow if the scaled result exceeds the type's range.
   ```
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -4,7 +4,7 @@ The single source of truth for coding style and conventions in this repository.
 It serves humans and AI agents alike: contributors follow it by hand, and every
 Dev3 stage and code-review tool reads it instead of restating the rules.
 
-> Scope: **how** we write Move in this repo — naming, ordering, idioms, testing,
+> Scope: **how** we write Move in this repo - naming, ordering, idioms, testing,
 > documentation. For **why** the library is shaped the way it is (design
 > decisions, package layout, object model), see [`ARCHITECTURE.md`](./ARCHITECTURE.md).
 
@@ -16,7 +16,7 @@ These conventions target **Move 2024** (module label syntax, receiver syntax,
 
 - **Error constants**: `EPascalCase`, declared with `#[error(code = N)]`, type
   `vector<u8>`, plain `"..."` string literal (NOT `b"..."`). Codes are sequential
-  from 0 and **append-only** — never renumber existing errors when adding new
+  from 0 and **append-only** - never renumber existing errors when adding new
   ones. Callers (frontends, integrators, tests in other packages) may match on
   the numeric abort code; renumbering silently breaks them. New errors go at the
   bottom with `code = N+1`.
@@ -27,17 +27,17 @@ These conventions target **Move 2024** (module label syntax, receiver syntax,
 - **CRUD**: `new`/`create`, `empty`, `borrow`/`borrow_mut`, `add`/`remove`,
   `exists`/`contains`, `drop`/`destroy`/`destroy_empty`, `to_name`/`from_name`
 - **Dynamic field keys**: positional struct with `Key` suffix
-- **Generic type parameters**: `T`, `U` — use descriptive names only when they
+- **Generic type parameters**: `T`, `U` - use descriptive names only when they
   materially aid readability
 - No "Potato" in hot-potato struct names
 
 ## Module & Package
 
-- `module my_package::my_module;` — no braces, 2024 edition label syntax
+- `module my_package::my_module;` - no braces, 2024 edition label syntax
 - One module per object/data structure
 - PascalCase package names, snake_case named addresses; prefix named addresses
   with the project name (`openzeppelin_*`)
-- Since Sui 1.45, Sui / MoveStdlib / Bridge / SuiSystem are implicitly imported —
+- Since Sui 1.45, Sui / MoveStdlib / Bridge / SuiSystem are implicitly imported -
   no manual dep declaration needed
 
 ## Section ordering
@@ -48,12 +48,12 @@ Every file must use `// === <Name> ===` delimiters in this order:
 2. `// === Errors ===`
 3. `// === Constants ===`
 4. `// === Structs ===`
-5. `// === Events ===` — placed **after** Structs when both are present in the
+5. `// === Events ===` - placed **after** Structs when both are present in the
    same module; in a dedicated events module, this is the only struct-family
    section and Structs is omitted
-6. `// === Method Exports ===` (if any) — `public use fun ...` receiver-syntax
+6. `// === Method Exports ===` (if any) - `public use fun ...` receiver-syntax
    aliases exposed by this module
-7. `// === Init ===` (if an `init` function is present — place it first)
+7. `// === Init ===` (if an `init` function is present - place it first)
 8. `// === Public Functions ===`
 9. `// === View helpers ===`
 10. `// === Admin Functions ===` (if any)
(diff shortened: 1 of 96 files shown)


bidzyyys:
@ericnordelo https://move-book.com/guides/code-quality-checklist/#do-not-prefix-tests-with-test_-in-testing-modules

STYLEGUIDE.md
```diff
@@ -233,13 +238,22 @@ Examples of required rewrites:
 
 ## Documentation
 
+Doc-comments are the source of truth for the generated API reference
+(`sui move build --doc`) and for downstream AI-integrator tooling — treat them as
+part of the public API, not decoration. Keep them complete and accurate.
+
```
ericnordelo, line 244: Nice!

contracts/access/tests/access_control_tests.move
```diff
@@ -63,7 +63,7 @@ fun take_ac(scenario: &Scenario): AccessControl<ACCESS_CONTROL_TESTS> {
 
 #[test]
 #[allow(lint(share_owned))]
-fun test_new_with_otw_succeeds() {
```
ericnordelo, line 70: Do we have to remove this? Similar to removing test_only from the module definition, I think it may harm readability. I think this prefix helps to quickly recognize that a function is a tests, for example, when you are ctrl+shift+F through the workspace, similar to why keeping the tests suffix in the file name helps with readability even when is not a requirement. cc @0xNeshi for you thoughts too.

bidzyyys: https://move-book.com/guides/code-quality-checklist/#do-not-prefix-tests-with-test_-in-testing-modules

0xNeshi: I agree with Daniel's choice here.

Ctrl+shift+F-ing is straightforward even without the prefix, as you have many indicators you're looking at a test function: `#[test]` attribute, `_tests` suffix in the file name, name describes a flow instead of an action (and usually includes `succeeds/fails`). 

I'd go as far as removing the `_tests` suffix from test module names (like is done in Rust), but having that suffix is a convention in Move on Sui (aligns test module name with the file name).

STYLEGUIDE.md
```diff
@@ -233,13 +238,22 @@ Examples of required rewrites:
...
 - `///` for doc comments (renders in IDEs), `//` for inline technical notes
 - No JavaDoc-style `/** */`
 - Document struct fields, complex params, and return values
 - Use section headings `#### Parameters`, `#### Returns`, and `#### Aborts` when
   relevant; use `-` for list items (not `*`)
-- Document public functions with at least `Parameters` and `Returns`; include an
-  `Aborts` section whenever a function can abort
+- Document public functions: include `#### Parameters` and `#### Returns` where
+  they add detail beyond the one-line summary — a trivial getter whose summary
```
0xNeshi:
```suggestion
  they add detail beyond the one-line summary - a trivial getter whose summary
```
Update all em dashes into regular dashes, see https://github.com/OpenZeppelin/contracts-sui/pull/388

STYLEGUIDE.md
```diff
@@ -233,13 +238,22 @@ Examples of required rewrites:
...
+  already states what it returns needs neither. Always include an `#### Aborts`
+  section whenever a function can abort, listing **every** error it can raise —
+  including errors propagated from internal calls
```
0xNeshi: Apply this rule to all existing functions, i.e. double check if all docs follow this rule. 

Also, align the following docs:
- in `math/core` none of the `fun inv_mod` and `fun mul_mod` implementations list the actual underlying abort error, e.g. see [u16.move](https://github.com/OpenZeppelin/contracts-sui/blob/645efdcf1a7621a44fc677ab11385cbdf468b110/math/core/sources/u16.move#L197-L215) and [u256.move](https://github.com/OpenZeppelin/contracts-sui/blob/645efdcf1a7621a44fc677ab11385cbdf468b110/math/core/sources/u256.move#L205-L223)
- in `macros.move`, in the [fun mul_div](https://github.com/OpenZeppelin/contracts-sui/blob/645efdcf1a7621a44fc677ab11385cbdf468b110/math/core/sources/internal/macros.move#L169-L171), the format of the `#### Aborts` message is misaligned with existing convention

0xNeshi: Specify should the `#### Aborts` section appear if no custom errors are emitted.
Two options:
1. No section - any native errors (e.g. arithmetic) are implied
2. Display section, but only if native errors can appear

Currently there are functions that don't follow any established convention:
- [macros.move > mul_shr](https://github.com/OpenZeppelin/contracts-sui/blob/645efdcf1a7621a44fc677ab11385cbdf468b110/math/core/sources/internal/macros.move#L169-L171) -> specifies no custom errors are emitted
- [macros.move > checked_shl](https://github.com/OpenZeppelin/contracts-sui/blob/645efdcf1a7621a44fc677ab11385cbdf468b110/math/core/sources/internal/macros.move#L98-L101) -> specifies that native errors can appear

STYLEGUIDE.md
```diff
@@ -233,13 +238,22 @@ Examples of required rewrites:
...
+- State caller preconditions explicitly and map each to the error it fails with
+  (e.g. "Aborts with `EUnauthorized` if the caller lacks the role")
```
0xNeshi:
```suggestion
  (e.g. "`EUnauthorized` if the caller lacks the role")
```
This is the actual current convention, see [access_control.move](https://github.com/OpenZeppelin/contracts-sui/blob/645efdcf1a7621a44fc677ab11385cbdf468b110/contracts/access/sources/access_control.move#L428-L430).
Related to this, align the abort message format of [decimal_scaling.move > validate_decimals](https://github.com/OpenZeppelin/contracts-sui/blob/645efdcf1a7621a44fc677ab11385cbdf468b110/math/core/sources/decimal_scaling.move#L218-L220).

contracts/access/tests/access_control_tests.move
ericnordelo: This one is the one that has the less strong argument to keep, so I'm ok with removing it. But to me is not the same as removing test_only annotations inside test files, or removing the test suffix on test files, in the sense that there is no an easy visual way then of recognizing these elements without checking the path to see where they are contained, for example when you do ctrl+P on cursor, or when you ctrl+shft+F. It looks like a regression to me on those cases, since there is no clear advantage besides removing a few short lines of code.

0xNeshi: I'm ok with leaving `test_` prefix. It doesn't improve clarity for me, but I don't mind leaving it if it helps you.

STYLEGUIDE.md
bidzyyys: Done in afc170a — the em dashes in the new prose are now regular dashes (this got picked up during the main sync). The only intentional `—` left is the example on the rule line itself ("instead of em dashes (`—`)").

STYLEGUIDE.md
bidzyyys: Done in afc170a:
- `inv_mod` / `mul_mod` wrappers (u8…u256) now name the concrete error: `` - `EZeroModulus` if `modulus` is zero. `` instead of the generic "Aborts if …".
- `macros.move > mul_div`: `#### Aborts` reformatted to the convention — `` - `EDivideByZero` if `$denominator` is zero. ``
- Also named `EDivideByZero` in the `mul_div` wrappers, and swept the whole `math/core` for other generic `Aborts if …` / colon-style bullets — none remain.

bidzyyys: Picked option 2: the STYLEGUIDE `#### Aborts` bullet now says to list **every** cause including native aborts (arithmetic overflow, division by zero). So the section is present iff the function can actually abort — native causes included.

Applied in afc170a:
- `mul_shr`: removed its `#### Aborts` section — it cannot abort (overflow is returned as a flag, and `shift: u8` is always a valid shift amount), so "Does not emit custom errors" was misleading.
- `checked_shl` / `checked_shr`: kept their sections — they *can* abort natively when `shift >= bit-width`, which the doc already states.

STYLEGUIDE.md
bidzyyys: Adopted the suggestion in afc170a — the precondition example now follows the actual `` - `Error` if … `` convention from access_control.move (dropped the "Aborts with" prefix).

Also aligned `decimal_scaling.move`: `validate_decimals` plus the two colon-style bullets (`EInvalidDecimals` / `ESafeDowncastOverflowedInt`) at L57/58/131 are now all `` - `Error` if … ``.

contracts/access/tests/access_control_tests.move
bidzyyys: @ericnordelo @0xNeshi that the feedback PR we got in the past from Mysten Labs on v1.0.0: https://github.com/OpenZeppelin/contracts-sui/commit/724611928077952de989c17d790bc6dd4a5eab4e#diff-d79a241ddd260e0a0756f4aa3212561ce3f0430d5e9013855d9922d27b3a0f86

TL;DR they removed `test_*` prefix.

0xNeshi (approved):
Excellent work

ericnordelo (approved):
LGTM!

PR #395, opened 2026-06-17
feat: Vesting Wallet

### LINKED ISSUE #156: [Feature]: Implement Vesting Wallet

Implement a VestingVallet based on #155.

0xNeshi:
Resolves #156 

Performed the following validity checks:
- `/sui-btt`
- `/sui-basic-review`
- Ran Ultracode workflow with the sole task of exploiting the vesting wallet by releasing unvested funds - the implementation passed the audit :heavy_check_mark: 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog

## Summary by CodeRabbit

* **New Features**
  * Added vesting wallet functionality for managing beneficiary asset vesting with flexible, customizable schedules.
  * Added linear vesting schedule implementation with support for cliffs and stepped tranches.

* **Documentation**
  * Added comprehensive test coverage specifications and invariant documentation.
  * Added formal review and coverage assessment reports.


contracts/finance/README.md
```diff
@@ -0,0 +1 @@
+# `openzeppelin_finance`
```
0xNeshi, line 1: `vesting-wallet-artifacts/*` is a temporary directory used for local development, will be removed prior to merge

contracts/finance/sources/linear_schedule.move
```diff
@@ -0,0 +1,235 @@
+/// The built-in linear-with-cliff schedule for `vesting_wallet` - a reference curve
```
0xNeshi: I'm not 100% certain on the name for this module.
The name `linear_schedule` is correct in the sense that it provides schedule calculation logic. However, it also provides convenience wrapper functions to interact with the underlying `VestingWallet`, so it's not just a "schedule logic" module.
A more accurate name would be `vesting_wallet_linear` 🤔 

By extension, I'd add additional schedules with convenience functions, e.g. `vesting_wallet_stepped`

0xNeshi: In the financial world, `linear` is just a synonym for "graded", which entails a "stepped" curve as well.
See definitions, they're the same:
- [linear vesting](https://ledgy.com/blog/vesting-schedules)
- [graded vesting](https://www.investopedia.com/terms/g/graded-vesting.asp)

We can update linear to represent both these curves. I could then add 2 more curves:
1. Graded - release tokens in unequal tranches (e.g. `10/20/30/40` or `20/30/20/30`) -> https://github.com/OpenZeppelin/contracts-sui/pull/399
2. Milestone-based - release tokens on specific milestones achieved

bidzyyys:
### Review follow-ups

Code itself looks solid — authorization model, fund conservation, monotonicity and the overflow guards all check out, and the 62 tests pass locally. Remaining items, mostly author decisions:

- [x] **Wire the package into CI** — done in caa75f7: `contracts/finance` added to the CI package matrix (`.github/workflows/test.yml`) + a note in `CONTRIBUTING.md` so new packages get registered in the same PR. Until this, the package was never built/linted/tested in CI.
- [x] **Fix the `E99011` lint now surfaced by CI** — `Lint contracts/finance` fails on `vesting_wallet_linear_tests.move:28`: `teardown(test: Scenario, clk: Clock)` takes `Clock` by value (`uncallable_function`). Resolve it (take `&Clock` / restructure / `#[allow(lint(uncallable_function))]`). NB: `contracts/utils/tests/rate_limiter_tests.move:15` has the identical pattern yet passes — worth understanding the difference.
- [x] **`Move.toml` is missing the `[addresses]` section** — sibling packages declare `[addresses] openzeppelin_<pkg> = "0x0"`; `contracts/finance/Move.toml` only has `[package]`. Builds today (Move.lock resolves `root`), but it's inconsistent with the repo convention.
- [x] **`vesting-wallet-artifacts/` committed at repo root** — these are skill-pipeline process docs (`invariants.md`, `tests.md`, `post-core/basic-review.md`, `btt-coverage.md`); no other package ships such a directory. Likely should not be part of the PR. (This is also where CodeRabbit's INV-12 discrepancy lives.)
- [x] **Package docs** — `contracts/finance/README.md` is just the `# openzeppelin_finance` heading; the module-level doc-comments are excellent, but the README is empty and the "Documentation"/"Changelog" checklist boxes are unchecked.

Discussion (non-blocking):
- [ ] **Naming: `Linear` vs "stepped"** — the `Linear` witness/module actually implements a stepped/tranche curve (linear is the `period_ms = 1` limit). The docs explain it well; just flagging the name↔concept mismatch for awareness.

bidzyyys:
### Review follow-ups (from `/sui-basic-review`)

Additional items from an independent basic review — not repeating the earlier list. Code itself is clean: no Critical/High, all 7 security categories pass, every inferred invariant is enforced.

- [x] **Root cause of the `E99011` lint = framework-rev drift in `Move.lock`** (answers my earlier "worth understanding the difference" note). The byte-identical `teardown(test: Scenario, clk: Clock)` passes in `contracts/utils` but fails in `contracts/finance` purely because the pinned Sui framework revs differ — finance `718ae563…` (stricter `uncallable_function`) vs utils `c2428b3a…` vs access `655576ca…`. Preferred fix is to **unify the pinned Sui rev across all packages** (regenerate the lock so finance matches the repo), which clears the lint without touching the test. The per-package rev divergence is also a latent hygiene issue worth fixing regardless.
- [x] **Permissionless `destroy` can strand in-flight deposits** — once `clock >= end` and balance is zero, anyone may `destroy`; a `Coin` already `public_transfer`'d to the wallet address but not yet `receive_and_deposit`'d is then stranded, and the storage rebate goes to the arbitrary destroyer. It's documented and `ENotEnded` guards against front-running *future* funding, but consider either beneficiary-restricting teardown or explicitly accepting the tradeoff.
- [x] **Minor test gaps** (non-blocking) — continuous schedule with `cliff_ms == duration_ms` boundary isn't tested (the stepped equivalent is); and `vesting_wallet::release` driven through a third-party wrapper (the documented wrapper use case) is covered only indirectly.

bidzyyys:
### Review follow-ups (from `/sui-btt`)

BTT coverage walk against the source-derived tree. Coverage is tight — every `assert!` failure path has an `#[expected_failure]` test, all four events assert full payloads, both topologies and composability flows are exercised. Three genuine (minor) gaps, none duplicating earlier comments:

- [x] **Zero-value `deposit` boundary** (Medium) — `deposit` with a `Coin` of value 0 passes the overflow guard and emits `Deposited { amount: 0 }`. No leaf pins this. Decide intent first: should a zero deposit short-circuit (no event), or is emitting `Deposited{0}` intended? The answer makes this either a missing guard or a missing test. Branch: `deposit / when coin value == 0 / it {emits Deposited{0} | no-ops}`.
- [x] **Owned-topology `receive_and_deposit`** (Low) — only the shared-wallet receive path is tested (`receive_and_deposit_claims_addressed_coin` shares the wallet first). The documented "owned fast path" — holder claims a coin sent to the wallet's address — isn't exercised directly. One owned-mode receive test closes it.
- [x] **Linear `releasable` view at pre-start / pre-cliff** (Low) — the view returning 0 before the schedule opens is covered only transitively through `release`; assert `vesting_wallet_linear::releasable(...) == 0` directly at a pre-start clock and a pre-cliff clock.

contracts/finance/sources/vesting_wallet_linear.move
```diff
@@ -0,0 +1,442 @@
...
+        let total = wallet.balance() + wallet.released();
+        // SAFETY: construction guarantees `period_ms * steps` and`start_ms + period_ms * steps`
+        // fit in u64, so neither arithmetic here overflows.
+        if (now >= start_ms + period_ms * steps) {
+            total
+        } else {
+            let elapsed_steps = (now - start_ms) / period_ms;
+            // SAFETY: `now < start_ms + period_ms * steps`, so `elapsed_steps < steps`:
+            // the staircase value stays strictly below `total` until the post-end clamp.
+            mul_div(total, elapsed_steps, steps)
```
bidzyyys: Shouldn't you use `openzeppelin_math::mul_div`?

bidzyyys (approved):
Looks good to me 👍

contracts/finance/sources/vesting_wallet.move
```diff
@@ -0,0 +1,609 @@
...
+/// #### Parameters
+/// - `wallet`: The wallet to destroy. Must hold a zero balance.
+///
+/// #### Returns
+/// - A `DestroyReceipt<S, P>` carrying the wallet's beneficiary and schedule
+///   parameters, to be passed to `consume_receipt`.
+///
+/// #### Aborts
+/// - `ENotEmpty` if the wallet still holds a balance.
+public fun destroy_empty<S: drop, P: copy + drop + store, C>(
```
0xNeshi, line 434: # Why `destroy_empty` is permissionless instead of witness-gated

**TL;DR:** if tearing down a `VestingWallet` required the curve witness `S`, a
curve-agnostic wrapper could never destroy a wallet it holds - because it
doesn't have, and can't fabricate, `S`.

## The reasoning

- A witness `S` can only be minted inside its own curve module. Any code outside
  that module - including a generic wrapper that just *holds*
  `VestingWallet<S, P, C>` without knowing the curve - has no way to produce one.
- So if `destroy_empty` took `_w: S`, the only caller that could ever destroy the
  wallet is the curve module itself. A wrapper that owns the wallet would be
  stuck: it can move it, but never tear it down and reclaim the storage rebate.
  That breaks curve-agnostic composability, which is the whole point of letting
  wrappers hold these.

## The fix (#404): split the authority via a hot-potato receipt

- `destroy_empty` takes **no witness** → permissionless. Any holder (incl. a
  wrapper) can drain-and-destroy the wallet and get back a `DestroyReceipt<S, P>`.
- `consume_receipt` is **witness-gated** (`_w: S`) → only the declaring curve can
  unwrap the receipt.
- The receipt is a hot potato, so it *must* be consumed in the same PTB. That
  drags the curve into the teardown transaction regardless of who initiated it,
  and lets the curve run teardown logic and **veto by aborting** (which reverts
  the whole PTB, including the destruction).
- This mirrors the existing `VestedAmount` split: one half callable without the
  witness, the curve gates the other. Same authority model, consistent across
  the API.

**Net:** wrappers get to destroy, the curve keeps its veto. Witness-gating the
destroy directly would have given the curve veto power but at the cost of making
curve-agnostic teardown impossible.

contracts/finance/README.md
```diff
@@ -0,0 +1,276 @@
...
+        30 * DAY_MS,  // period_ms
+        4,            // steps
+        ctx,
+    );
+    wallet.deposit(funds);
+    transfer::public_share_object(wallet);
+}
+
+// Continuous one-year linear vest with a 90-day cliff. `create_and_share` only
+// covers the stepped form, so share the continuous wallet explicitly.
```
ericnordelo: Is this true? Can't we use create and share with period_ms = 1 instead?

contracts/finance/sources/vesting_wallet.move
```diff
@@ -0,0 +1,609 @@
...
+/// - **Shared** (recommended): `transfer::public_share_object(wallet)` - anyone
+///   can poke `release`. `vesting_wallet_linear` exposes `create_and_share` sugar.
+/// - **Owned** (fast path): `transfer::public_transfer(wallet, addr)` - only the
+///   holder can pass the wallet by `&mut`, so funding and release are reachable
+///   from the holder's transactions only. Outside parties fund it by
+///   `public_transfer`ing a `Coin<C>` to the wallet's object address; the holder
+///   then claims each with `receive_and_deposit`.
+///
+/// The `beneficiary` is fixed at construction (mirroring OpenZeppelin's
+/// `VestingWallet`). To rotate the recipient, point `beneficiary` at a
```
ericnordelo: The mirroring OpenZeppelin's VestingWallet part is not accurate, as OZ uses Ownable for it. I would just remove this parenthesis phrase.

ericnordelo (commented):
Looking great @0xNeshi . Left just a couple of minor comments.

contracts/finance/sources/vesting_wallet_linear.move
```diff
@@ -0,0 +1,482 @@
...
+/// - `clock`: Sui `Clock`, used to check the schedule has ended.
+/// - `ctx`: Transaction context, used to check the caller is the beneficiary.
+///
+/// #### Aborts
+/// - `ENotEnded` if called before the schedule's end (`start_ms + period_ms * steps`).
+/// - `ENotBeneficiary` if the caller is not the wallet's beneficiary.
+public fun destroy(receipt: DestroyReceipt<Linear, Params>, clock: &Clock, ctx: &mut TxContext) {
+    let (beneficiary, params) = vesting_wallet::consume_receipt(receipt, Linear {});
+    assert!(clock.timestamp_ms() >= params.calculate_end(), ENotEnded);
+    assert!(ctx.sender() == beneficiary, ENotBeneficiary);
```
0xNeshi, line 380: I'm not fully convinced on this design choice. It does protect from stranded coins, but it limits the design integrators must use (e.g. see [pausable_grant::unwrap](https://github.com/OpenZeppelin/contracts-sui/pull/400/changes#diff-4fa862bea1e2aca9b6687175ff3f86d7f71eafdd4b7e7ad174c8f73bc9654537R160)).

Two alternatives I've considered:
1. anyone can destroy -> reintroduces the "stranded coin" exploit (if we can even call it that)
2. on wallet creation, a `DestroyCap` is minted, which the caller can choose whether to share or who to send it to -> introduces an additional step on wallet creation, but it makes the linear curve more flexible, and makes the integrators fully in charge of the destruction process.

@ericnordelo @bidzyyys

ericnordelo: What is the "stranded coin" exploit?

ericnordelo: Why do you think unwrap is problematic or limiting? Isn't it better to converge into the same flow by just unwrapping?

0xNeshi: > What is the "stranded coin" exploit?

see [docs](https://github.com/OpenZeppelin/contracts-sui/blob/5c7229b39a31db55c7726a3d9f6027d5a7a18dfa/contracts/finance/sources/vesting_wallet_linear.move#L352-L359):
> /// Both extra gates guard against stranding an in-flight deposit. The ended gate stops
/// a wallet being torn down ahead of a pending deposit, front-running funding intended
/// to arrive later. The beneficiary gate addresses the residual case: a coin
/// `public_transfer`'d to the wallet's address but not yet `receive_and_deposit`'d is
/// invisible to `destroy_empty`'s empty check, so - since `destroy_empty` is
/// permissionless - an arbitrary actor could otherwise strand such a deposit and
/// pocket the storage rebate. Restricting this final step to the beneficiary keeps both
/// the strand risk and the rebate with the only party harmed by it.

---

> Why do you think unwrap is problematic or limiting? Isn't it better to converge into the same flow by just unwrapping?

Unwrap isn't the problem, the `ctx.sender() == beneficiary` assert is. The assert forces the unwrapping pattern, and makes it impossible to delete a wallet if the beneficiary was set to an object (e.g. [splitter example](https://github.com/OpenZeppelin/contracts-sui/blob/690de2960f00cce05c15a2bca9d6b7087b3dbf98/contracts/finance/examples/vesting_wallet/splitter.move#L40)).

contracts/finance/README.md
0xNeshi: It is no longer true, I added a continuous version for `create_and_share_continuous`, see https://github.com/OpenZeppelin/contracts-sui/pull/395/commits/5c7229b39a31db55c7726a3d9f6027d5a7a18dfa

0xNeshi: will update

contracts/finance/sources/vesting_wallet_linear.move
ericnordelo: I suggest we merge with the current design, and raise the discussion points explicitly at the start of the audit, with alternatives provided ina a very concise, short, but descriptive way for auditors to understand it as quickly as possible, and then we decide if we keep it or refactor it as part of the audit, by continuing the internal devs discussion with auditors input. Trying to iterate on this today will just delay the audit and I think theres no need.

ericnordelo (approved):

PR #399, opened 2026-06-18
feat: add graded schedule

0xNeshi:
Towards #156 

Opened as separate PR from https://github.com/OpenZeppelin/contracts-sui/pull/395 for easier and focused review just on the schedule logic.


contracts/finance/sources/vesting_wallet_graded.move
```diff
@@ -0,0 +1,328 @@
...
+
+        if (bps == 0) {
+            0
+        } else {
+            // SAFETY: depositing has a check ensuring no balance overflow can occur.
+            let total = wallet.balance() + wallet.released();
+            // SAFETY: `bps <= BPS_DENOMINATOR` (construction caps it at `10_000`), and
+            // `mul_div` uses a u128 intermediate, so `total * bps` cannot overflow. At
+            // the last stage `bps == BPS_DENOMINATOR`, so this clamps exactly to `total`.
+            mul_div(total, bps, BPS_DENOMINATOR)
```
bidzyyys, line 312: Why not `openzeppelin_math::mul_div`?

bidzyyys (approved):
Missing updates in README etc.

bidzyyys:
### Review — `vesting_wallet_graded` (after the branch update)

Curve itself is correct and clean, parallel to `vesting_wallet_linear`: construction validation is solid (non-empty, equal lengths, strictly-increasing offsets and bps, final bps `== 10_000`, `start + last_offset` overflow guard), and `vested_amount_raw` is monotonic and bounded by `total` (`mul_div`, bps ≤ 10_000, u128 intermediate, last stage == exactly `total`). No cliff param is an intentional, documented choice. Local check: package builds, 103 tests pass, strict lint (`--lint --warnings-are-errors`) and docgen clean.

- [ ] **`destroy` is missing the beneficiary gate that `linear::destroy` now has** (Medium / consistency) — after this branch picked up the updated base, the two sibling curve modules have different teardown authorization: `linear::destroy(wallet, clock, ctx)` asserts `ENotEnded` **and** `ENotBeneficiary` (`ctx.sender() == beneficiary`), while `graded::destroy(wallet, clock)` asserts only `ENotEnded`. This leaves the exact gap that was closed in `linear`: anyone can destroy an ended+empty graded wallet, stranding an in-flight `receive_and_deposit` coin and pocketing the storage rebate. Align it: add `ctx: &mut TxContext` + an `ENotBeneficiary` error + the `ctx.sender() == wallet.beneficiary()` assert, and update the doc (it still carries the old "destruction is permissionless" rationale).
- [ ] **`graded_tests::teardown(test, clk: Clock)` reintroduces the by-value-`Clock` helper** (Low / consistency) — #395 deliberately removed this pattern from `linear_tests` (it tripped the `uncallable_function` lint). It currently lints clean here, but align with the base convention (destroy the clock at the call site / drop the helper) so the package stays uniform.
- [ ] **No explicit upper bound on stage count** (Informational) — `offsets_ms`/`cumulative_bps` are unbounded vectors stored in the wallet object. Only the caller's own wallet is affected (bounded in practice by the 250 KB object limit), and `linear` doesn't bound its step count either, so this is a "consider it" note, not a defect.

The curve logic, accounting, and monotonicity/bound invariants all check out — the only action item is aligning `destroy` (and its test helper) with the `linear` sibling.

bidzyyys:
### `/sui-basic-review` — `vesting_wallet_graded`

No Critical/High/Medium **bug**. The module is a thin, correct curve over the `vesting_wallet` primitive; all inferred invariants hold. The one actionable item is the `destroy` consistency gap (also raised above).

**Invariant verification (inferred) — all ✅**
- Curve monotonic non-decreasing in time — `vested_amount_raw` scans to the highest elapsed stage; offsets & cumulative_bps strictly increasing (enforced at `new`).
- Curve bounded by `balance + released` — `mul_div(total, bps, 10_000)` with `bps ≤ 10_000`; last stage `bps == 10_000` ⇒ exactly `total`, so the honest curve never trips `EInsufficientBalance`.
- Schedule well-formed — `new` enforces non-empty, equal lengths, strictly-increasing offsets (index 0 may be 0), strictly-increasing positive bps, final `== 10_000`, and `start + last_offset` u64-overflow guard.
- Witness-gated mint/teardown — `Graded` is `drop`-only, module-private; `Params` module-private.
- Pre-start / pre-first-stage vest zero — early returns.

**Security checklist**
- Access control — PASS (mint/destroy witness-gated; release/deposit permissionless by primitive design). **Caveat:** `destroy` is ended-only, not beneficiary-gated like `linear::destroy` — see the Medium item above.
- Arithmetic — PASS (`mul_div` u128 intermediate; bps capped at 10_000; construction overflow-guarded; floor rounding, dust released at the final 100% stage — no funds lost).
- Type safety — PASS (phantom `Graded`/`C`, `Params` stored).
- Reentrancy/composability — PASS (release CEI + idempotent, inherited from primitive; `total` re-derived each call so retroactive deposits participate).
- Object safety / economic / upgrade — PASS / N/A (no value minted; curve only attests).

**Verdict:** Ready once `destroy` is aligned with `linear` (Medium item above). Curve math and accounting are sound.

bidzyyys:
### `/sui-btt` — `vesting_wallet_graded` test coverage

**Tight**, parallel to the `linear` suite (31 graded tests; package builds, 103 tests pass, strict lint clean). The module emits no events of its own (they come from the primitive), so the leaf set is curve-shape + construction guards + lifecycle.

**Covered (`✅`)**
- All 7 construction/teardown error codes have `#[expected_failure]`: `EEmptySchedule`, `ELengthMismatch`, `EUnsortedOffsets`, `EInvalidBps` (×3), `EIncompleteSchedule`, `EScheduleOverflow`, `ENotEnded` — plus primitive `EBalanceOverflow` and `ENotEmpty`.
- Curve shape: `vested_amount_pre_start_is_zero`, `vested_amount_is_a_graded_staircase` (per-stage values), `vested_amount_post_end_clamps_to_total`, `vested_amount_is_nondecreasing_in_time` (monotonicity), `new_accepts_single_stage` (pure-cliff boundary), `release_before_first_stage_moves_no_funds`.
- Release/lifecycle/composability: stage-portion release + permissionless, within-stage re-release no-op, releasable-matches-release, full-release-after-end, retroactive-deposit-never-over-releases, overflowing-refund-rejected, create+deposit+release in one flow.

**Candidate gaps**
- [ ] **u128-intermediate at max** (Low) — `linear` has `vested_amount_uses_u128_intermediate_at_max`; graded has no equivalent worst-case (`total = u64::MAX`, a mid stage with `bps < 10_000`) test pinning that `mul_div` doesn't overflow. Add one for parity.
- [ ] **multi-stage catch-up jump** (Low) — a clock that jumps past several stages at once should land on the highest elapsed stage's cumulative bps. The scan covers it and the staircase test steps through, but an explicit "skip 2+ stages in one step" assertion would pin it directly.
- [ ] **destroy-with-undrained-pool strand** (Low / by-design) — same documented teardown footgun as the other curves; likely impractical to assert post-UID-deletion in `test_scenario` (Deferred if so).

No `❌` on any invariant-backed leaf; the gaps are parity/boundary nice-to-haves.

0xNeshi:
Closing as not planned for now

PR #411, opened 2026-06-25
ci: split fast tests from coverage tracing

immrsd:
## What

Splits the single `test` job into two, per @0xNeshi's review:

- **`test`** — `sui move test` (no tracing): the fast regression gate (~5s even for `math/core`). Gates merges.
- **`coverage`** — `sui move test --trace` + `sui move coverage lcov` + Codecov: isolated, `continue-on-error: true` (informational, never blocks a merge), runs only after `test` is green.

## Why the 30-minute bump wasn't enough

Profiled `math/core` locally (Sui 1.71.1):

- `sui move test` (no trace): **~5.5s**.
- `sui move test --trace` (what CI runs): the single random test `median_matches_sorted_reference_u256` alone takes **7m21s and writes 4.1 GB** of traces. It `quick_sort!`s a Sui-generated random `vector<u256>` that can be very large, and `--trace` (full mode) records every VM instruction. The whole suite is many GB and never finishes inside 30 min.

`--trace` (full) is mandatory for `sui move coverage lcov` — the lighter `--trace function-only` / `instruction-only` modes produce no line records — so the cost can't simply be switched off.

## The fix

@0xNeshi's split keeps the gate fast and moves the slow trace off the critical path. The coverage job also passes `--rand-num-iters 1`: coverage only needs each line executed once, and the full randomized property iterations still run in the fast `test` job, so this cuts trace volume without losing coverage.

## Deferred follow-ups (`math/core`'s domain)

- The durable cure for the trace time is bounding the random vector length in `math/core`'s ~9 heavy `median`/`quick_sort` random tests (one of them produces the 4.1 GB above). Left out here since it changes those property tests.
- `traces/` is not gitignored — worth adding so a local `--trace` run can't accidentally commit gigabytes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Split automated checks into two stages for a faster primary test run and a separate coverage run.
  * Kept the main test job on a shorter timeout while allowing coverage collection to run independently.
  * Adjusted coverage execution to reduce trace output and improve reliability of test reporting.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -103,7 +103,7 @@ jobs:
     name: Test ${{ matrix.package }}
     needs: [setup, lint]
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
     permissions:
       contents: read
       pull-requests: write
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,14 +99,13 @@ jobs:
         run: sui move build --doc
         working-directory: ./${{ matrix.package }}
 
+  # Fast, untraced run: the primary regression gate. Finishes in seconds even
+  # for the heaviest suites, so a failing test is reported quickly.
   test:
     name: Test ${{ matrix.package }}
     needs: [setup, lint]
     runs-on: ubuntu-latest
-    timeout-minutes: 30
-    permissions:
-      contents: read
-      pull-requests: write
+    timeout-minutes: 15
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -103,7 +103,7 @@ jobs:
     name: Test ${{ matrix.package }}
     needs: [setup, lint]
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
```
0xNeshi, line 138: If this is really the cause of the issue, then nice catch! 

I'm not sure about increasing the timeout globally though. 
Since all other tests finish way below the 15 minute mark, this increase would potentially waste CI time on a test suite that hangs for an actually valid reason! 
We also still have a problem of not quickly being notified on whether a regression was introduced into `math/core` without waiting on the whole trace to be created.

I suggest the following update: **create a separate `sui move test` job (no tracing) that _executes fast_. That way we quickly realize if there are regressions. The `sui move test --trace` job still runs, but as a separate focused job, and it has this increased timeout.**

immrsd: Good call - implemented the split.

Profiling `math/core` locally showed why even 30 min wasn't enough: `sui move test` is ~5s, but under `--trace` the single random test `median_matches_sorted_reference_u256` alone is **7m21s / 4.1 GB** - it sorts a large random `vector<u256>` and full tracing records every VM instruction. The lighter `--trace` modes (`function-only`/`instruction-only`) don't produce lcov line data, so tracing can't just be made cheaper.

So now:
- `test` runs `sui move test` (no trace) as the fast gate.
- `coverage` runs `--trace` + lcov separately, `continue-on-error`, only after `test` passes, with `--rand-num-iters 1` (coverage needs each line once; the property iterations stay in `test`).

The real cure for the trace time is bounding those random vector lengths in `math/core`'s heavy tests - left that out here since it touches the property tests, but happy to follow up.

0xNeshi (approved):

ericnordelo (approved):

PR #414, opened 2026-06-25
feat: Implement Token Sale

0xNeshi:
Resolves #319 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog

## Summary by CodeRabbit

* **New Features**
  * Added a fixed-rate, prefunded token sale for Sui with full lifecycle phases, buyer-bound purchase receipts, refund escrow handling, optional vesting redemptions, and admin-controlled emergency cancellation.
  * Introduced allowlist-gated participation and a fixed-rate pricing curve to compute quotes and allocations.

* **Documentation**
  * Added comprehensive README coverage for the new sale package and updated the changelog with an Unreleased entry.

* **Tests**
  * Added extensive unit and integration tests for setup, purchases, refunds, claims, lifecycle transitions, allowlist, pricing, and receipt behavior.

* **Chores**
  * Updated coverage ignore rules.


bidzyyys:
## Review — sale

- [x] 🟡 **prefunded_sale.move:847** (`share_and_activate`) — the module doc guarantees "buyer claims and refunds do not depend on admin liveness" (L48), but the paired `RefundVault` is shared only by a separate integrator step (`refund_vault::share`) that the library neither performs nor verifies. Activate without sharing the vault and it stays *owned by the issuer*, so the permissionless `cancel_after_close` / `refund` paths (which need `&mut vault`) become uncallable by buyers — silent in the success path (`claim` needs no vault), locking buyer funds only on a soft-cap miss → have `share_and_activate` take the `RefundVault<PaymentCoin>` by value, assert `object::id(vault) == *refund_vault_id.borrow()`, and `share_object` it, so the no-admin-liveness guarantee is structural rather than convention.

contracts/sale/sources/prefunded_sale.move
```diff
@@ -0,0 +1,2232 @@
...
+
+    let (mut wallet, destroy_cap) = vesting_wallet::new<Witness, VestingScheduleParams, SaleCoin>(
+        *sale.vesting_schedule_params.borrow(),
+        ctx.sender(), // only buyer can claim
+        ctx,
+    );
+    wallet.deposit(payout);
+
+    (wallet, destroy_cap)
+}
```
ericnordelo, line 1563: H-1 — A buyer can bypass the vesting schedule entirely (demonstrated).

claim_into_vesting<…, Witness: drop> lets the caller pick the vesting witness type. Nothing binds that Witness to the curve module that owns the sale's stored schedule params. In openzeppelin_finance, the witness is the only gate on minting a vested-amount attestation — so a buyer supplies a drop witness declared in their own package, mints a full attestation, and releases everything immediately. I proved it with a PoC: on a 1-year sale with a 90-day cliff, the buyer released their full allocation at activation + 1 ms. This defeats the issuer-imposed lockup — a headline feature — and directly contradicts "the buyer cannot influence or bypass the schedule." Bounded to the buyer's own allocation (no theft), so High, not Critical.

Fix: pin the witness — store type_name::get<Witness>() in set_vesting_schedule_params and assert it in the two vesting-claim paths (or make the vesting witness a sale type parameter).

contracts/sale/sources/prefunded_sale.move
```diff
@@ -0,0 +1,2232 @@
...
+///   Consumed.
+/// - `clock`: Sui `Clock`, read for the current timestamp.
+///
+/// #### Aborts
+/// - `ETicketSaleMismatch` if `ticket` was minted for a different sale.
+/// - `ENotInit` if the sale is not in `Init` phase.
+/// - `EVaultRequiredForActivate` if no refund vault has been paired.
+/// - `EActivationAfterClose` if `now >= closes_at_ms`.
+/// - `EInsufficientInventoryAtActivate` if `inventory < required_inventory`.
+public fun share_and_activate<
```
ericnordelo, line 1035: H-2 — Activating with an un-shared refund vault bricks the permissionless close/refund paths. (This is the same issue @bidzyyys already flagged.)

share_and_activate shares the sale but not the vault, and never verifies the vault gets shared — that's a separate integrator step. If it's skipped, the vault stays owned by the issuer, and every close/refund path (finalize, cancel_after_close, refund) needs &mut vault, which only the owner can supply. So "permissionless" finalize/cancel_after_close become issuer-only, and on a soft-cap miss buyer funds sit in the vault recoverable only if the issuer later shares it — permanently locked if the issuer is gone. Contradicts "buyer redemption never depends on admin liveness" (stated 3×). It's silent on the success path and only bites at cancel.

Fix: make sharing atomic — have share_and_activate take the vault by value and share it alongside the sale.

ericnordelo (commented):
Took a look to the PR and design while running a review with Claude. In general looks good @0xNeshi. Left a couple of comments on critical things, and leaving a list of things here to review:

Medium

- M-1: purchase is all-or-nothing against the hard cap — no partial fill. Near sell-out, a payment that would exceed hard_cap reverts entirely; buyers must size to remaining capacity off-chain. Likely intended, but undocumented — decide + document.

- M-2: withdraw_proceeds / withdraw_unsold_inventory have no "already drained" guard: a second call returns an empty balance but still emits a zero-amount event. Add a zero-guard (as refund_vault::deposit already does).
Informational

- I-1: The PrefundedSale struct bound VestingScheduleParams: copy is looser than the copy + drop + store used everywhere else. Harmless (the only constructor forces the full set — I verified even a bare generic compiles), but misleading; align it. This is exactly where CodeRabbit posted a 🔴 Critical "does not have 'store'" — that's a false positive on current HEAD, since the package compiles and all tests pass.

- I-2: prefunded_sale_curve_trust_tests.move:26 has a local setup() duplicating test_utils::setup() — valid CodeRabbit nitpick; drop it.

- I-3: No teardown path for a fully-redeemed sale (shared forever; contributions table persists) — consistent with the "stale receipts pin funds" design, worth an explicit note.

bidzyyys:
## Review — sale

- [x] ⚪ **prefunded_sale.move:11-29** — the lifecycle diagram uses non-ASCII box-drawing glyphs (`──`, `│`, `├`, `┴`, `└`) and `×N`, but STYLEGUIDE mandates regular ASCII characters in all docs and the sibling `refund_vault.move:10-17` diagram is already ASCII-only → redraw with `-` / `|` / `+` / `-->` / `xN` to match.
- [x] ⚪ **prefunded_sale.move:1174, 1222** — `cancel_after_close` / `cancel_emergency` `#### Aborts` list `phase::EAlreadyCancelled`, but no `phase` module or such error exists; `do_cancel` sets `Phase::Cancelled` unconditionally with no assertion → remove the bogus propagated-abort bullet (doc-vs-code drift).
- [x] ⚪ **refund_vault.move:45-62 & prefunded_sale.move:146+** — `#[error(code = N)]` values are grouped in tens (`0,1,2,10,20…`), but STYLEGUIDE requires codes "sequential from 0"; these are all-new errors so renumbering is safe (tests reference by name) → renumber sequentially, or confirm the category-grouping is an intentional deviation.

contracts/sale/Move.toml
bidzyyys, line 1: Shouldn't be part of `openzeppelin_finance`? Wdyt @0xNeshi @ericnordelo ?

bidzyyys:
## Review — sale (re-review of current head)

Correctness/invariants re-verified — all core invariants (refund & claim solvency, phase monotonicity, vault-state coupling, per-buyer cap, vesting-witness pinning, inventory backing) hold; H-1, H-2, M-1, M-2, I-1, I-3 from earlier rounds are addressed. Only style/consistency nits remain:

- [x] ⚪ **fixed_rate_curve.move:69, 122, 160** — section labels diverge from the other four sale modules and STYLEGUIDE: `// === Constructors ===` / `// === Quote ===` are used with no `// === Public Functions ===` umbrella, and `// === Views ===` should be `// === View helpers ===` (the canonical name the other modules use) → align them. (Extends the still-open CodeRabbit section-naming nit, which didn't cover the `Views` label.)
- [x] ⚪ **prefunded_sale_claim_refund_tests.move:32, 44** — `#[test_only]` on `finalize_now` / `cancel_now` is redundant inside a `tests/` module (every item there is already test-only), and the `// allows accepting &mut Clock` comment is incorrect (a plain private helper accepts `&mut Clock` fine) → drop both attributes and the comment.
- [x] ⚪ **prefunded_sale.move:2217** (`is_open`) — compares `sale.phase != Phase::Active` directly while every other phase check goes through the `is_active()` / `is_*()` predicates → use `!sale.phase.is_active()` for consistency.

0xNeshi:
re: [2nd point](https://github.com/OpenZeppelin/contracts-sui/pull/414#issuecomment-4927605641) -> linter complains if `#[test_only]` isn't present

contracts/sale/Move.toml
0xNeshi: When we were creating rate limiter, @ericnordelo had an [argument](https://openzeppelin.slack.com/archives/C08HUPCCS7Q/p1781533873096819?thread_ts=1781530585.596129&cid=C08HUPCCS7Q) on creating as many separate packages as possible:

Though now we're not publishing the packages, so this argument might make no sense now? :thinking: 

I'm fine with either approach.

ericnordelo: I still think we should split them, in case someone wants to publish one and not the other

ericnordelo:
Getting there, one last thing

- N-1 (Low, docs): the README usage examples weren't updated for the new 6th type parameter. The source doc-comments were updated, but the README's Move examples still show 5 type args and would not compile:

create_sale<FixedRateCurve, FrcParams, SALE, USDC, VParams> (line 219) and the PrefundedSale<…> in the buy example (line 259) are missing VestingWitness.
The "Redeem into vesting" section (lines 305–313) still says the witness is "the final type argument" of claim_into_vesting and passes <…, VParams, Linear> — but after the fix the witness is chosen at create_sale and inferred by claim_into_vesting, so the prose is wrong and the arg order is swapped.
Since the README is the primary integration guide, I'd fix this before merge — regenerate the examples against the 6-param API and rewrite the vesting section.

Also


- M-1 no partial-fill at hard cap | ⏳ Open — untouched, still undocumented
- I-3 no teardown path | ⏳ Open by design

0xNeshi:
@ericnordelo re: https://github.com/OpenZeppelin/contracts-sui/pull/414#issuecomment-4928473522

> * N-1 (Low, docs): the README usage examples weren't updated for the new 6th type parameter. The source doc-comments were updated, but the README's Move examples still show 5 type args and would not compile:

Thanks, addressed.
 
> * M-1 no partial-fill at hard cap | ⏳ Open — untouched, still undocumented

This is actually addressed, see https://github.com/OpenZeppelin/contracts-sui/commit/9d733c46303c9e3ea6f8442ffef9d7a62f3a2ceb

> * I-3 no teardown path | ⏳ Open by design

Correct, this is open by design.

bidzyyys:
## Review — sale (prefunded_sale, refund_vault, allowlist, fixed_rate_curve, receipt)

Critical pass: strict conventions + invariants + BTT. Build/lint clean both modes, 111/111 tests. Fund-conservation, phase completeness, overflow guards, allowlist, refund-vault state machine, and vesting witness-pinning all verified sound. The prior review set (H-1 witness bypass, H-2 unshared-vault brick, M-1 partial-fill doc, M-2 zero-amount withdraw event, I-1 loose `copy` bound) all verify as fixed on HEAD. Net-new items only:

### Test coverage (BTT)
- [x] 🟡 **prefunded_sale_purchase_tests.move:87** — `Purchased` is the only event asserted by `.length() == 1` instead of full-payload equality; `test_new_purchased` (prefunded_sale.move:2540) is defined but never used → capture the delivered receipt id and assert `events[0] == test_new_purchased(paid, allocation, raised_after, ...)`, matching every other event.
- [x] 🟡 **window boundary `now == closes_at_ms` untested** — `purchase` is inclusive (`now <= closes_at_ms`, L1165) but `finalize`/`cancel_after_close` are strict (`now > closes_at_ms`, L1269/L1335); no test sets the clock to `CLOSES` (5_000) → add: (a) `purchase` at `now == closes_at_ms` succeeds, (b) non-sold-out `finalize` at `now == closes_at_ms` still aborts `ESaleWindowStillOpen`. Guards the `<=`-vs-`>` off-by-one.
- [x] ⚪ same-code/different-branch + view gaps: `refund`'s own `EReceiptSaleMismatch` (L1764) only tested via the `claim` site; `claim_all` `EClaimRequiresVesting` (L1481) and `claim_all_into_vesting` `ENoVestingScheduleAttached` (L1607) untested; `is_open` (L2199, 3 branches) and `has_reached_soft_cap` (L2230) unexercised; `cancel_emergency` success at `now == closes_at_ms` and with a configured-but-unmet soft cap (2nd disjunct of L1394) untested.

### Docs
- [x] ⚪ **prefunded_sale.move:1225-1226** — `finalize` doc says it closes the vault "so the admin can later withdraw proceeds", but on the success path the vault is never funded and `withdraw_proceeds` (L1659) pulls from `sale.proceeds`; `refund_vault::withdraw_all` is unreachable (cap is trapped in the sale). The flip is lifecycle hygiene, not a proceeds path → reword to avoid sending integrators looking for a vault-side withdrawal.

ericnordelo (approved):
LGTM!

contracts/sale/README.md
```diff
@@ -0,0 +1,430 @@
...
+| [`fixed_rate_curve`](https://docs.openzeppelin.com/contracts-sui/1.x/api/sale#fixed_rate_curve) | The built-in pricing curve: `allocation = paid * rate`, fixed for the whole sale. Mints the `Quote` and `ActivationTicket` a `FixedRateCurve` sale needs. **Most sales use this.** |
+| [`refund_vault`](https://docs.openzeppelin.com/contracts-sui/1.x/api/sale#refund_vault) | A generic refundable escrow over `Balance<P>`. Every sale is paired with one; on cancel it holds the proceeds and pays buyers back individually. Usable standalone. |
+| [`allowlist`](https://docs.openzeppelin.com/contracts-sui/1.x/api/sale#allowlist) | A typed compliance slot (`AllowlistAdmin` + single-use `AllowEntry`). The library ships **no** KYC logic - you wire your own scheme against these types. |
+| [`receipt`](https://docs.openzeppelin.com/contracts-sui/1.x/api/sale#receipt) | The non-transferable, buyer-bound claim ticket minted by `purchase` and consumed by `claim` / `refund`. |
+| [`phase`](https://docs.openzeppelin.com/contracts-sui/1.x/api/sale#phase) | The shared lifecycle enum: `Init -> Active`, then terminal `Finalized` or `Cancelled`. |
```
0xNeshi:
```suggestion
```

0xNeshi:
The [formatter CI failure](https://github.com/OpenZeppelin/contracts-sui/actions/runs/29344581281/job/87124864389?pr=414) is peculiar, and it seems to be a `main` code/CI pipeline issue

contracts/sale/sources/fixed_rate_curve.move
```diff
@@ -0,0 +1,187 @@
...
+///
+/// ### Purchase
+///
+/// The buyer threads `quote + purchase` in the same PTB. `quote` takes
+/// the payment as a `Balance` (the `Quote` carries it through to
+/// `purchase`); `purchase` consumes the quote and delivers a `Receipt`
+/// to the buyer:
+///
+/// ```move
+/// let quote = fixed_rate_curve::quote(&sale, payment.into_balance());
```
bidzyyys, line 34: Why not `receiver.syntax()`? I mean: 
- [ ] `sale.quote(payment.into_balance());`
- [ ] `sale.purchase(quote, allow, &clock, ctx);`

contracts/sale/README.md
```diff
@@ -0,0 +1,429 @@
...
+> is out of scope for this package.
+
+## Install
+
+```toml
+[dependencies]
+openzeppelin_sale = { r.mvr = "@openzeppelin-move/sale" }
+```
+
+> [!NOTE]
```
bidzyyys, line 32: It needs to be adjusted later.

bidzyyys (approved):
LGTM

contracts/sale/sources/fixed_rate_curve.move
0xNeshi: Because `fixed_rate_curve::quote` and `sale` are from different modules. We'd need to define a local alias to use the method syntax

contracts/sale/README.md
0xNeshi: what do you mean?

0xNeshi: updated the dependency to use the git format

PR #429, opened 2026-07-02
Gaussian math: `inverse_cdf`

### LINKED ISSUE #251: [Feature]: `inverse_cdf` for fixed-point math library

Follow the [reference](https://paragraph.com/@evandekim/on-chain-atomic-gaussian-math).

immrsd:
## Summary by CodeRabbit

* **New Features**
  * Added inverse CDF (standard-normal quantile) support for signed and unsigned fixed-point types, including domain handling for probabilities and upper-half evaluation with reflection.
  * Added deterministic saturation and symmetry behavior to match quantile expectations.

* **Documentation**
  * Documented the new `inverse_cdf` API, accuracy/behavior guarantees, and code generation workflow for quantile support.

* **Tests**
  * Added new unit tests and oracle-based test vectors covering accuracy, monotonicity, symmetry, round-trip behavior, and determinism.

* **Chores / CI**
  * Extended code generation validation in CI and added inverse-CDF coefficient/vector drift checks.


bidzyyys:
- [x] 🟡 **inverse_cdf_coefficients.move:67 + `sd29x9_base`/`ud30x9_base` inverse_cdf docs** — `MAX_Z_RAW = 6_300_000_000` and the doc claims "6.3 matches the CDF domain bound, so the two functions agree at the corner". PR #431 moves the CDF saturation to `6.109410205`, so once both land that claim is false and `cdf`/`inverse_cdf` no longer agree at the corner (inverse_cdf saturates to 6.3 while cdf already treats `≥6.109` as 1.0). Reconcile: either align inverse_cdf saturation to the new CDF bound, or document the intentional divergence.

bidzyyys (approved):
lgtm

immrsd:
Good catch — will fix this in #431 rather than here:

- On current `main` the claim still holds: cdf's bound is 6.3 today.
- Aligning now would break the corner instead of fixing it — today's cdf returns 0.999999999 at z = 6.109410205, so `cdf(inverse_cdf(1))` would stop being 1 until #431 lands.
- #431 already has to touch `inverse_cdf` (new `horner_eval!` parameter), so it will move the sentinel to 6.109410205 there (constants regenerated, fit unchanged) and add a test asserting `inverse_cdf_coefficients::max_z_raw() == cdf_coefficients::max_z_raw()` and `cdf(inverse_cdf(1)) == 1` — so the two bounds can't drift apart again.

ericnordelo (approved):
LGTM!

PR #444, opened 2026-07-06
[N-07] Timelock audit

ericnordelo:
## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration updates that don’t actually change a value no longer emit extra change events.
  * Timelock operations still complete normally, but redundant updates are now silent.

* **Tests**
  * Added coverage for no-op updates to confirm the expected events are emitted only when values change.

* **Documentation**
  * Clarified event behavior so it’s clear that reapplying the current setting does not produce an event.


bidzyyys (approved):
LGTM

contracts/timelock/sources/timelock.move
```diff
@@ -667,7 +671,9 @@ public fun execute_update_min_delay<Role>(
 
     let previous_ms = self.min_delay_ms;
     self.min_delay_ms = new_min_delay_ms;
-    event::emit(MinDelayChanged { previous_ms, new_ms: new_min_delay_ms });
+    if (previous_ms != new_min_delay_ms) {
```
0xNeshi, line 674: Consider moving this check above all mutations, to avoid wasting gas on a no-op.

```move
    if (previous_ms == new_min_delay_ms) {
        return;
    }

    // ...
    assert!(new_min_delay_ms <= MAX_DELAY_MS, EInvalidConfig);
    ...
```

Same for other functions

ericnordelo: I don't think we gain much from this small change, as it will only save self.min_delay_ms, since we still need execute_internal. Let's keep it as is for simplification, keeping in mind we need to deliver fast.

0xNeshi (approved):

PR #446, opened 2026-07-06
[N-01] VestingWallet audit

ericnordelo:
## Summary by CodeRabbit

* **Documentation**
  * Clarified wrapper usage for wallet/vault examples to show that address-targeted funding must be forwarded through the wrapper.
  * Added guidance on re-exposing the claim-and-deposit flow so funds sent to an inner wallet remain accessible and do not appear stranded.


bidzyyys (approved):
LGTM

PR #449, opened 2026-07-06
[N-04] VestingWallet audit

ericnordelo:
## Summary by CodeRabbit

* **Documentation**
  * Clarified how scheduled releases behave when vested amounts temporarily move backward, including when releases stop versus when they continue with a smaller payout.
  * Updated the vesting schedule notes to reflect the integer precision used in stepped-curve calculations.


bidzyyys (approved):
LGTM

PR #451, opened 2026-07-06
[N-06] VestingWallet audit

ericnordelo:
## Summary by CodeRabbit

* **Documentation**
  * Clarified vesting release behavior, including that it can be triggered by anyone, may be called repeatedly as vesting progresses, and always preserves the total amount.
  * Added guidance that payouts may arrive as multiple small coins over time, so integrators should be prepared to handle many incoming transfers for object-based beneficiaries.


bidzyyys (approved):
LGTM

PR #454, opened 2026-07-07
feat: add collections package (sorted_map, sorted_set)

### LINKED ISSUE #118: [Feature]: Set/Sorted Set

Set implementation for Sui Ecosystem.

Issue raised during Consumer Roundtable on 4th December 2025.

### LINKED ISSUE #261: [Feature]: SkipList

Implement a generic `SkipList`. Use [Cetus move-stl](https://suivision.xyz/package/0xe93247b408fe44ed0ee5b6ac508b36325b239d6333e44ffa240dcc0c1a69cdd8?tab=Code) as a reference.

kosedogus:
Resolves #261
Resolves #118

The two ordered-collection modules ship as a single `openzeppelin_collections`
package (and one PR) since they will be audited together (`sorted_set` is a
thin wrapper over `sorted_map`).

Line coverage: `sorted_map` and `sorted_set` 100%.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog

## Summary by CodeRabbit

* **New Features**
  * Added `openzeppelin_collections` with ordered `sorted_map` and `sorted_set`, including comparator-aware ordering, pagination, and navigation utilities.
  * Added Move examples: sorted-map order book, prize vault, tick registry; sorted-set allowlist, unlock queue, validator set.
* **Documentation**
  * Added collections package documentation and updated repository layout/catalog pointers, architecture guidance, and changelog.
* **Tests**
  * Added extensive sorted-map and sorted-set unit, differential, type-safety, comparator-contract, abort-safety, and value-conservation tests.
* **Chores**
  * Expanded CI package matrix for collections and increased coverage timeout.


collections/examples/sorted_map/order_book.move
```diff
@@ -0,0 +1,152 @@
...
+// === Structs ===
+
+/// Aggregate resting size at one price.
+public struct Level has copy, drop, store {
+    /// Total resting size aggregated at this price level.
+    size: u64,
+}
+
+/// A two-sided book embedding two `SortedMap`s - the only state the integrator owns.
+public struct OrderBook has key {
```
0xNeshi, line 42: Every example uses `u64` as the map key. Would be useful to have an example where `key` is a struct, like in `sorted_set`'s [Validator](https://github.com/OpenZeppelin/contracts-sui/pull/454/changes#diff-ac059d646f9a1b0c05a006566e3653d791c07c08486d7e221bfa88cf90b236d0R38).

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,877 @@
+/// A generic, ordered key->value map backed by a single sorted `vector<Entry<K, V>>`.
+///
+/// `SortedMap` is a UID-less value: no identity of its own, no dynamic fields. Every entry
+/// lives inline in one vector, shaped like `sui::vec_map::VecMap`. A bare map is not `key`, so
+/// embed it in your own `has key` object for owned or shared semantics:
+/// ```
```
0xNeshi:
```suggestion
/// ```move
```
nit: add the language identifier to code blocks

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,880 @@
...
+/// position.
+///
+/// Abilities materialize jointly over `K` and `V`: `Entry<u64, u64>` is `copy + drop +
+/// store`, while `Entry<u64, Coin<T>>` is store-only.
+public struct Entry<K: copy + drop + store, V: store> has copy, drop, store {
+    key: K,
+    value: V,
+}
+
+/// A map kept sorted by key, backed by one contiguous `vector<Entry<K, V>>`.
```
0xNeshi: I would refrain from documenting internal implementation. If the fact that `vector` is used as the backing data structure is important, then the structure itself should be renamed to `sorted_vec_map::SortedVecMap`, thus being self-documenting.

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,880 @@
...
+/// A pure value - no `UID`, no dynamic fields - so it embeds directly in an integrator's
+/// object, exactly like `sui::vec_map::VecMap`. `copy`/`drop` materialize only when both
+/// `K` and `V` allow them: `SortedMap<u64, u64>` is `copy + drop + store`;
+/// `SortedMap<u64, Coin<T>>` is store-only and must be drained then `destroy_empty`'d.
+public struct SortedMap<K: copy + drop + store, V: store> has copy, drop, store {
+    /// The map's entries. Across the supported (macro) API this vector is strictly
+    /// increasing under the (consistently supplied) comparator: sorted, with no duplicate
+    /// keys. The forced-public position-based writers (`insert_at`/`remove_at`) can break
+    /// this if misused - see the forced-public internals note in the module header.
+    entries: vector<Entry<K, V>>,
```
0xNeshi: Did you consider using [VecMap](https://docs.sui.io/references/framework/sui_sui/vec_map#sui_vec_map_VecMap) as the underlying data structure? The current implementation seems like a reimplementation of `VecMap`

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,880 @@
...
+    value: V,
+}
+
+/// A map kept sorted by key, backed by one contiguous `vector<Entry<K, V>>`.
+///
+/// A pure value - no `UID`, no dynamic fields - so it embeds directly in an integrator's
+/// object, exactly like `sui::vec_map::VecMap`. `copy`/`drop` materialize only when both
+/// `K` and `V` allow them: `SortedMap<u64, u64>` is `copy + drop + store`;
+/// `SortedMap<u64, Coin<T>>` is store-only and must be drained then `destroy_empty`'d.
+public struct SortedMap<K: copy + drop + store, V: store> has copy, drop, store {
```
0xNeshi:
```suggestion
public struct Entry<K: copy, V> has copy, drop, store {
    key: K,
    value: V,
}

/// A map kept sorted by key, backed by one contiguous `vector<Entry<K, V>>`.
///
/// A pure value - no `UID`, no dynamic fields - so it embeds directly in an integrator's
/// object, exactly like `sui::vec_map::VecMap`. `copy`/`drop` materialize only when both
/// `K` and `V` allow them: `SortedMap<u64, u64>` is `copy + drop + store`;
/// `SortedMap<u64, Coin<T>>` is store-only and must be drained then `destroy_empty`'d.
public struct SortedMap<K: copy, V> has copy, drop, store {
```
Why do we need to restrict abilities so much? Sui's [VecMap](https://github.com/MystenLabs/sui/blob/2d157da8e85274f19ec2a78eb6da39156736a5d5/crates/sui-framework/packages/sui-framework/sources/vec_map.move#L30) uses the suggested ability set just fine.

Currently `K` requires `drop` only so that [remove_at](https://github.com/OpenZeppelin/contracts-sui/pull/454/changes#diff-68f046139b6d70a6b72ef43e441b635fa64d841a4a1d5b9c67af861d8410c659R243) could return just the value. In fact, it would make sense for `remove_at(i)` to return the key + value pair (since the caller might not know the `key` of the index they removed). 
Even better, `remove_at(i)` should return `Entry<K, V>`, as tuples `(K, V)` can't be wrapped in `Option`. This would make the existing API that returns `Option`s fine, where `Option<V>` would be turned into `Option<Entry<K, V>>`.
Devs would still have a way to access keys and values using `entry_key` and `entry_value`.
Consider this less restrictive design.

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,880 @@
...
+    entries: vector<Entry<K, V>>,
+}
+
+// === Public Functions ===
+
+// === Lifecycle ===
+
+/// Create a new, empty map. Takes no `&mut TxContext`: a `SortedMap` is a value, not an
+/// object.
+public fun new<K: copy + drop + store, V: store>(): SortedMap<K, V> {
```
0xNeshi: Missing `#### Returns` section

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,880 @@
...
+    SortedMap { entries: vector[] }
+}
+
+/// A map holding a single `key`/`value` entry. O(1); needs no comparator, since one entry is
+/// trivially sorted. For more than one entry from pre-sorted data, use `from_sorted_keys_values!`.
+public fun singleton<K: copy + drop + store, V: store>(key: K, value: V): SortedMap<K, V> {
```
0xNeshi: Missing returns and parameters sections. Double check other public functions.

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,880 @@
...
+public fun assert_key_found(found: bool) {
+    assert!(found, EKeyNotFound);
+}
+
+/// Abort `EUnequalLengths` if `equal` is false. Routed through this regular fun so the bulk
+/// constructor's abort fires at this module's location, not the consumer's inlined macro body.
+///
+/// #### Aborts
+/// - `EUnequalLengths` if `equal` is false.
+public fun assert_equal_lengths(equal: bool) {
```
0xNeshi, line 337: Used in only 1 place, inline the assertion

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,880 @@
...
+    assert!(equal, EUnequalLengths);
+}
+
+/// Abort `EKeysNotStrictlyIncreasing` if `increasing` is false. Routed through this regular fun
+/// so the abort fires at this module's location, not the consumer's inlined macro body.
+///
+/// #### Aborts
+/// - `EKeysNotStrictlyIncreasing` if `increasing` is false.
+public fun assert_strictly_increasing(increasing: bool) {
```
0xNeshi, line 346: Used in only 1 place, inline the assertion

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,880 @@
...
+    &mut map.entries.borrow_mut(i).value
+}
+
+/// Abort `EKeyNotFound` if `found` is false. Routing absent-key lookups through this
+/// regular fun guarantees the abort fires at this module's location, not in the
+/// consumer's inlined macro body.
+///
+/// #### Aborts
+/// - `EKeyNotFound` if `found` is false.
+public fun assert_key_found(found: bool) {
```
0xNeshi, line 318: Used in only 2 places, inline the assertion

collections/sources/sorted_map.move
kosedogus: They are stricter than `vec_map` deliberately. `SortedMap` will be the leaf for the df backed `BigSortedMap` and the two will be bridged by `from_sorted_map` and `into_sorted_map` functions. `BigSortedMap` will be pinned to `K: copy+drop+store, V:store` by df + B-tree reabalancing. So its framework analogs are `Table` and `LinkedTable` not `vec_map`. Relaxing `SortedMap` this way would make it a superset of what `BigSortedMap` can hold, so we would lose 1-to-1 bridging guarantee.
Also, `remove` returning `Entry` part: It diverges from `Table`'s `remove`, it would need an `Entry` destructor we need to expose, and would also add friction to integrators (for example `prize_vault` gets its coin back directly). Extra generality also seems not much reachable in practice, every read returns `K` by copy (so copy is required regardless), and we wouldn't be able to embed a non-store value anway.
Let me know if I am missing something though, this is a very important decision that definitely needs to be handled correctly.

collections/sources/sorted_map.move
kosedogus: I wouldn't call it a reimplementation. `VecMap` is insertion-ordered not sorted, has O(n) membership, it has no insert-at-position and no binary search ofc. Its `contents` vector is private so you can't keep it sorted or get a cheaper lookup + ordered navigation on top of it. These are the things we are adding with `SortedMap`.

collections/tests/sorted_map/abort_tests.move
```diff
@@ -0,0 +1,217 @@
...
+// === borrow / borrow_mut on an absent key -> EKeyNotFound ===
+
+#[
+    test,
+    expected_failure(
+        abort_code = openzeppelin_collections::sorted_map::EKeyNotFound,
+        location = openzeppelin_collections::sorted_map,
+    ),
+]
+fun borrow_absent_below_head() {
```
0xNeshi:
```suggestion
#[test, expected_failure(abort_code = sm::EKeyNotFound, location = sm)]
fun borrow_absent_below_head() {
```
Since this alias is used for the import, wouldn't this work?

collections/sources/sorted_map.move
kosedogus: Changing documentation related to these.

collections/sources/sorted_map.move
0xNeshi: Fair enough

collections/sources/sorted_map.move
kosedogus: Writing for this and others below:
Unfortunately we cant, these are used in macro's. This one used in `from_sorted_keys_values!` for example or the `assert_key_found` below used in `borrow_by!`. Macro body can't reference the module's error constant, if we just inline the assertion with our error inside macro, it fails to compile. (Of course we can drop the module error, but thats not best-practice)

collections/tests/sorted_map/abort_tests.move
kosedogus: Yepp, verified it works. I'll switch them thank you

collections/sources/sorted_map.move
0xNeshi: > `BigSortedMap` will be pinned to `K: copy+drop+store, V:store` by df + B-tree reabalancing. So its framework analogs are `Table` and `LinkedTable` not `vec_map`.

True for `BigSortedMap`, ok then.

> Relaxing `SortedMap` this way would make it a superset of what `BigSortedMap` can hold, so we would lose 1-to-1 bridging guarantee.

I don't see how we lose any guarantees. A dev using `BigSortedMap` is bound by its limitations, no matter what `BigSortedMap` uses underneath. Same way we can use `vector` as the underlying structure, despite the fact it has no ability limits.

> Also, `remove` returning `Entry` part: It diverges from `Table`'s `remove`,

This is not the right comparison. `SortedMap::remove_at` removes an entry by _index_, while [Table::remove](https://docs.sui.io/references/framework/sui_sui/table#sui_table_remove) removes it by _key_. When removing data from specific position in a collection, one expects to receive all the data at that position, which in this case includes the `key`.

> it would need an `Entry` destructor we need to expose, and would also add friction to integrators

**EDIT: no friction would be introduced if we change the return type as [suggested](https://github.com/OpenZeppelin/contracts-sui/pull/454/#discussion_r3550949149).**

Right, but it would be just `let (key, value) = entry.unwrap()` or something like that, and I think it's worth the added flexibility. 
CONS: one extra line to destruct the entry.
PROS: `SortedMap` accepts a wider range of keys.

> and would also add friction to integrators (for example `prize_vault` gets its coin back directly).

Not sure what you mean with this example. Would the coin be used as the `key`? Nevertheless, not sure how this is relevant for the design decision.

> Extra generality also seems not much reachable in practice, every read returns `K` by copy (so copy is required regardless),

That's right, this ability would exist either way.

> and we wouldn't be able to embed a non-store value anway.

Technically speaking, you could embed it, just not leave it embedded/stored across transactions. Removing `store` is safe in any case, the real limitation is the `drop`.

@ericnordelo pinging to get your input here as well, as this is definitely important to get right

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+// Everything in this section is `public` because Move 2024 macro hygiene requires every
+// symbol a macro body references to be public at the consumer's expansion site (plus
+// `entry_value`, which completes the `entries_ref` read surface). In particular
+// `insert_at`/`remove_at` write at a caller-given index with no ordering check, so calling
+// them directly can corrupt sorted order. Use the macro API (`insert!`, `remove!`, ...) instead.
+
+/// Immutable view of the backing vector. There is deliberately no `&mut`/owning
+/// counterpart, so bulk reordering or bulk value-destruction is unrepresentable through
+/// this surface.
+public fun entries_ref<K: copy + drop + store, V: store>(
```
0xNeshi:
```suggestion
public fun entries<K: copy + drop + store, V: store>(
```
the immutable reference is generally implied, the convention for getters being:
- `fun [field-name]` - returns immutable reference
- `fun [field-name]_mut` - returns mutable reference

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+/// Borrow an entry's key. Macro bodies must read keys through this (not `.key`), since
+/// the field is private at the expansion site.
+public fun entry_key<K: copy + drop + store, V: store>(e: &Entry<K, V>): &K {
+    &e.key
+}
+
+/// Borrow an entry's value. Unlike its neighbors in this section, no macro body references
+/// it - it is public as the value-reading complement to `entry_key`, completing the read
+/// surface of `entries_ref` (the `Entry` fields are private).
+public fun entry_value<K: copy + drop + store, V: store>(e: &Entry<K, V>): &V {
```
0xNeshi:
```suggestion
public fun value<K: copy + drop + store, V: store>(e: &Entry<K, V>): &V {
```
this should compile; there are no other `fun value` declarations, so should be fine to use the generic name

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+/// this surface.
+public fun entries_ref<K: copy + drop + store, V: store>(
+    map: &SortedMap<K, V>,
+): &vector<Entry<K, V>> {
+    &map.entries
+}
+
+/// Borrow an entry's key. Macro bodies must read keys through this (not `.key`), since
+/// the field is private at the expansion site.
+public fun entry_key<K: copy + drop + store, V: store>(e: &Entry<K, V>): &K {
```
0xNeshi:
```suggestion
public fun key<K: copy + drop + store, V: store>(e: &Entry<K, V>): &K {
```
this should compile; there are no other `fun key` declarations, so should be fine to use the generic name

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+/// Borrow an entry's value. Unlike its neighbors in this section, no macro body references
+/// it - it is public as the value-reading complement to `entry_key`, completing the read
+/// surface of `entries_ref` (the `Entry` fields are private).
+public fun entry_value<K: copy + drop + store, V: store>(e: &Entry<K, V>): &V {
+    &e.value
+}
+
+/// Construct an entry, consuming `key` and `value` by move (no copy, no implicit drop).
+/// Harmless until fed to `insert_at`.
+public fun make_entry<K: copy + drop + store, V: store>(key: K, value: V): Entry<K, V> {
```
0xNeshi:
```suggestion
public fun new_entry<K: copy + drop + store, V: store>(key: K, value: V): Entry<K, V> {
```

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+
+/// Insert `e` at index `i`, shifting later entries right. Order-corruption surface: a
+/// non-sorted `i` leaves the vector unsorted.
+///
+/// #### Parameters
+/// - `i`: Insertion index.
+///
+/// #### Aborts
+/// - Native out-of-bounds abort inside `std::vector` if `i > length`.
+public fun insert_at<K: copy + drop + store, V: store>(
```
0xNeshi: Allowing custom index insertion seems out of place for a sorted map. I'd completely remove this function.

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+/// Remove the entry at index `i`, shifting later entries left, and return its value. The
+/// key is dropped; the value is moved out and returned, so no value is lost.
+/// Order-corruption surface if `i` is wrong.
+///
+/// #### Returns
+/// - The value at index `i`.
+///
+/// #### Aborts
+/// - Native out-of-bounds abort inside `std::vector` if `i >= length`.
+public fun remove_at<K: copy + drop + store, V: store>(map: &mut SortedMap<K, V>, i: u64): V {
```
0xNeshi: If access and removal at specific indices is important to expose, I think it would be more appropriate to rename the module and type to `sorted_vec_map::SortedVecMap`. That way it is clear why this map tracks indices.

Otherwise, I'd remove all of the `*_at(index)` functions, as that API makes no sense for a general map.

For clarity, I'm in favor of the module/type rename, but fine with either.

@ericnordelo

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+}
+
+/// `borrow_by` with the built-in integer `<`.
+///
+/// #### Returns
+/// - Reference to `key`'s value.
+///
+/// #### Aborts
+/// - `EKeyNotFound` if `key` is absent.
+public macro fun borrow<$K: copy + drop + store, $V: store>(
```
0xNeshi: Can't this be a regular function?

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+/// #### Parameters
+/// - `key`: Key to look up.
+/// - `lt`: Strict less-than comparator.
+///
+/// #### Returns
+/// - Reference to `key`'s value.
+///
+/// #### Aborts
+/// - `EKeyNotFound` if `key` is absent.
+public macro fun borrow_by<$K: copy + drop + store, $V: store>(
```
0xNeshi: Might be useful to have a `try_borrow` variant that returns `option::none()` if the key doesn't exist. This could help save gas in some use cases, since the dev wouldn't first have to check `SortedMap::contains` and then `borrow`, which both internally perform the search, but could just do a single search and match the option.

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,926 @@
...
+/// - `lt`: Strict less-than comparator.
+///
+/// #### Returns
+/// - `some(value)` if present (length - 1, order preserved), `none` if absent (map
+///   unchanged).
+public macro fun remove_by<$K: copy + drop + store, $V: store>(
+    $map: &mut SortedMap<$K, $V>,
+    $key: &$K,
+    $lt: |&$K, &$K| -> bool,
+): Option<$V> {
```
0xNeshi: Why does this return `Option`? Doesn't it make sense to abort if the key was not found?
Would make it possible to return a tuple `(K, V)` instead of `Entry<K, V>`, thus simplifying the removal of `drop + store` from the key type (no `entry.unwrap` would be necessary).

There's existing precedent for this:
- [vec_map::remove](https://github.com/MystenLabs/sui/blob/b76bcd0e2bc3551fda92dfff1eeaea75622061ab/crates/sui-framework/packages/sui-framework/sources/vec_map.move#L53)
- [table::remove](https://docs.sui.io/references/framework/sui_sui/table#sui_table_remove)
- [vector::remove](https://docs.sui.io/references/framework/sui_std/vector#std_vector_remove)
- Aptos's [ordered_map::remove](https://github.com/aptos-labs/aptos-core/blob/ee3597b87a5b36a8d89fe5665004130c8f0a308c/aptos-move/framework/aptos-framework/sources/datastructures/ordered_map.move#L131)

collections/sources/sorted_map.move
ericnordelo: After learning about [conditional abilities on generic types](https://move-book.com/reference/abilities/#conditional-abilities-and-generic-types) just now, which I haven't seen till now, I think we can safely relax the abilities on the generic types as @0xNeshi suggested, since it covers both key and values with and without store, which feel like both common use cases.

I don't see why it would affect integrations @kosedogus, but I unerstand that is hard to assess everything quickly. I recommend we pass the pattern doc to our LLM and ask it to check and refactor accordingly keeping integrations in mind, and we can improve it if needed later on audits.

0xNeshi: Integrations would be barely affected if we apply [this](https://github.com/OpenZeppelin/contracts-sui/pull/454/#discussion_r3550949149) `remove` change

collections/sources/sorted_set.move
```diff
@@ -0,0 +1,611 @@
...
+// symbol a macro body references to be public at the consumer's expansion site. They are
+// NOT a supported mutation API. `inner_mut` in particular lets a caller drive the wrapped
+// map directly with an inconsistent comparator or a wrong index and desort this set -
+// order-only corruption, local to that one set, no value ever lost. Use the macro API
+// (`insert!`, `remove!`, ...) instead.
+
+/// Immutable view of the wrapped map - what read macros (`contains!`, `find_*!`,
+/// `keys_from!`) expand against. Read-only: cannot be upgraded to `&mut`. Also the test
+/// order-check handle: `sorted_map::is_well_formed!(inner_ref(&set))`.
+public fun inner_ref<K: copy + drop + store>(set: &SortedSet<K>): &SortedMap<K, Unit> {
```
0xNeshi:
```suggestion
public fun inner<K: copy + drop + store>(set: &SortedSet<K>): &SortedMap<K, Unit> {
```
convention for ref getters

collections/sources/sorted_set.move
```diff
@@ -0,0 +1,611 @@
...
+    let mut set = new();
+    keys.do!(|k| { insert_by!(&mut set, k, $lt); });
+    set
+}
+
+/// `from_keys_by` with the built-in integer `<`. De-duplicates - see `from_keys_by`.
+///
+/// #### Returns
+/// - A set of the distinct keys, in ascending order.
+public macro fun from_keys<$K: copy + drop + store>($keys: vector<$K>): SortedSet<$K> {
```
0xNeshi:
```suggestion
public macro fun from_vector<$K: copy + drop + store>($elements: vector<$K>): SortedSet<$K> {
```

collections/sources/sorted_set.move
```diff
@@ -0,0 +1,611 @@
...
+/// reused expansion), never a straight-line sequence, to stay under Move's locals limit.
+///
+/// #### Parameters
+/// - `keys`: Keys to insert; duplicates are collapsed.
+/// - `lt`: Strict less-than comparator.
+///
+/// #### Returns
+/// - A set of the distinct keys, in comparator order.
+public macro fun from_keys_by<$K: copy + drop + store>(
+    $keys: vector<$K>,
```
0xNeshi:
```suggestion
public macro fun from_vector_by<$K: copy + drop + store>(
    $elements: vector<$K>,
```
Sets have no concept of "keys", that's a `SortedMap` concept.
I suggested `elements`, feel free to choose a different param name

collections/sources/sorted_set.move
```diff
@@ -0,0 +1,611 @@
...
+///
+/// #### Parameters
+/// - `keys`: Keys sorted (non-decreasing) under `lt`; compare-equal runs are collapsed.
+/// - `lt`: Strict less-than comparator.
+///
+/// #### Returns
+/// - A set of the distinct keys, in comparator order.
+///
+/// #### Aborts
+/// - `EKeysNotSorted` if `keys` has a strictly decreasing adjacent pair (not sorted under `lt`).
```
0xNeshi:
```suggestion
/// - `EKeysNotSorted` if `keys` has an adjacent pair not sorted under `lt`.
```
"decreasing" is the wrong term to use here (maybe user uses decreasing sorting order in `lt`, which is valid in their case), reword to sound more generic (don't have to use my suggestion).

collections/sources/sorted_set.move
```diff
@@ -0,0 +1,611 @@
...
+/// - `lt`: Strict less-than comparator.
+///
+/// #### Returns
+/// - `true` iff the key was newly added, `false` if it was already present.
+public macro fun insert_by<$K: copy + drop + store>(
+    $set: &mut SortedSet<$K>,
+    $key: $K,
+    $lt: |&$K, &$K| -> bool,
+): bool {
+    let set = $set;
```
0xNeshi:
```suggestion
public macro fun insert_by<$K: copy + drop + store>(
    $set: &mut SortedSet<$K>,
    $key: $K,
    $lt: |&$K, &$K| -> bool,
): bool {
    let set = $set;
    let key = &key;
```
applies to all macros: good practice is to assign the macro param to an internal variable

collections/sources/sorted_set.move
```diff
@@ -0,0 +1,611 @@
...
+/// - `key`: Key to remove.
+/// - `lt`: Strict less-than comparator.
+///
+/// #### Returns
+/// - `true` iff the key was present, `false` if it was absent.
+public macro fun remove_by<$K: copy + drop + store>(
+    $set: &mut SortedSet<$K>,
+    $key: &$K,
+    $lt: |&$K, &$K| -> bool,
+): bool {
```
0xNeshi: If we go ahead with making `Key` type non-drop, we should return the `$K` here.

collections/sources/sorted_set.move
```diff
@@ -0,0 +1,611 @@
...
+///
+/// #### Parameters
+/// - `from`: Lower-bound key.
+/// - `include`: Whether an exact match of `from` is included.
+/// - `limit`: Maximum number of keys to return.
+/// - `lt`: Strict less-than comparator.
+///
+/// #### Returns
+/// - Up to `limit` keys in ascending order.
+public macro fun keys_from_by<$K: copy + drop + store>(
```
0xNeshi: Not sure that sets have the concept of "keys". Consider a different name like `elements_from_by`

collections/sources/sorted_map.move
ericnordelo: I think sorted in the name already points to the idea of enumarability and indexing, mostly for the sake of delivering, I vote we keep the names and function as they are.

collections/examples/sorted_map/order_book.move
0xNeshi: We can add this later, no need to hold up the PR

collections/sources/sorted_set.move
0xNeshi: [sui::vec_set](https://docs.sui.io/references/framework/sui_sui/vec_set) uses the "keys" as a concept too, let's keep this for alignment

collections/sources/sorted_set.move
0xNeshi: [sui::vec_set](https://docs.sui.io/references/framework/sui_sui/vec_set) uses the "keys" as a concept too, let's keep this for alignment

collections/sources/sorted_map.move
ericnordelo: The precedents are strong and the API makes sense to me. Let's refactor it @0xNeshi

collections/sources/sorted_set.move
0xNeshi: [sui::vec_set](https://docs.sui.io/references/framework/sui_sui/vec_set) uses the "keys" as a concept too, let's keep this for alignment

collections/sources/sorted_set.move
0xNeshi: actually this is fine, because `$key` is passed to an internal macro

collections/sources/sorted_map.move
0xNeshi: can't, because of internal lambda

collections/sources/sorted_map.move
0xNeshi: Unfortunately, we need this for insertion within macros

bidzyyys:
## Review — collections

Build/lint clean both modes, 181/181 tests, all 6 abort codes covered; invariants (sorted-unique maintenance, `search!` lower-bound, resource-`V` conservation, dedup, reverse/coarse comparator, `keys_from` no-overflow) re-verified. One new quality nit not covered by the existing discussion:

- [x] ⚪ **sorted_set.move:183-187** (`singleton`) — reimplements insertion via the discouraged internal writer (`new()` + `set.inner.insert_at(0, sorted_map::make_entry(key, unit()))`) instead of delegating to the existing `sorted_map::singleton`: `SortedSet { inner: sorted_map::singleton(key, unit()) }` → simpler, avoids the "not-supported" `insert_at` path, and drops the field-access-vs-`inner_mut()` inconsistency with `from_sorted_keys_by`.

0xNeshi:
@ericnordelo @bidzyyys ready for review

collections/sources/sorted_map.move
0xNeshi: Can be implemented later

bidzyyys:
## Review — collections (sorted_map / sorted_set)

Strict pass on conventions + invariants + BTT. Build/lint clean both modes, 200/200 tests, BTT coverage complete, no correctness/invariant defects found. Three net-new nits only (everything else is already in the open discussion):

- [x] ⚪ **sorted_set.move:148-149** — `singleton` doc says the key "is placed at index 0 of the (empty) backing vector - the one provably order-safe direct `insert_at` the set performs", but the body now delegates to `sorted_map::singleton` (a `vector[Entry{..}]` literal, no `insert_at`); the set's only direct `insert_at` is in `from_sorted_keys_by` → reword to "constructed directly as a one-element map" and drop the `insert_at` claim.
- [x] ⚪ **sorted_set.move:69-70** — "Forced-public internals" enumerates `inner`, `inner_mut`, `unit` as exhaustive, but `assert_sorted` (L215) is also forced-public for the same macro-hygiene reason (referenced by `from_sorted_keys_by!` at L286) → add it, or hedge with "and similar" like the map header (L62).
- [x] ⚪ **sorted_set.move:86** — do NOT apply CodeRabbit's open "add `// === Imports ===` header" suggestion: no module in the repo uses that header (access_control, spend_vault, vesting_wallet, timelock, rate_limiter all have zero), and STYLEGUIDE lists item 1 "Imports" without a `// === ===` delimiter. Current code is already correct.

collections/sources/sorted_map.move
```diff
@@ -0,0 +1,969 @@
...
+/// - `EKeyNotFound` if `key` is absent.
+public macro fun remove_by<$K, $V>(
+    $map: &mut SortedMap<$K, $V>,
+    $key: &$K,
+    $lt: |&$K, &$K| -> bool,
+): $V {
+    let map = $map;
+    let (found, idx) = search!(map, $key, $lt);
+    assert_key_found(found);
+    let (_, value) = map.remove_at(idx);
```
ericnordelo: remove_by! cannot remove non-drop keys despite the advertised generic lifecycle. It discards the stored K, implicitly requiring K: drop, while its signature is unconstrained. SortedSet specifically tells users to drain non-drop keys through remove!, which is impossible. Provide a removal operation returning the stored key, preferably (K,V) for the map and K for the set, or accurately constrain and document the API.

collections/README.md
```diff
@@ -0,0 +1,177 @@
...
+- **Deliberate omissions.** No non-aborting `Option<&V>` borrow (Move cannot put a reference in `Option`) - use `if (contains!(..)) borrow!(..)`. No `values()`/`entries()` copy-out - `V` is only `store` (it may be a non-copyable resource like `Coin`), so values cannot be copied into a `vector`; read them per key via `borrow!`, or snapshot keys with `keys()` (or page with `keys_from!`) and `borrow!` each. No descending pagination - use a reverse comparator.
+
+### SortedSet
+
+- **The worst case is order-only - no key is ever lost.** Unlike the map (which can strand a `Coin` on misuse), the set's value is the trivial `Unit`, so a comparator violation gives wrong membership *answers* on a desorted set, but every key is still physically present and recoverable via `keys()`. This is the decisive simplification over the map. In tests, call `sorted_map::is_well_formed_by!(sorted_set::inner(&s), lt)` after `_by` sequences.
```
ericnordelo: The SortedSet security claim that no key is ever lost is false. Under a coarse comparator, upsert removes and drops the previous key, then stores the incoming byte-distinct key. The PR’s own test confirms {id:1, tag:100} disappears and only tag:200 remains. Moreover, is_well_formed_by! returns true in that test, so the recommended oracle cannot detect this comparator defect. The security notes materially understate the failure mode.

collections/examples/sorted_map/prize_vault.move
```diff
@@ -0,0 +1,155 @@
+/// A tournament prize pool that holds real coins in a `SortedMap` - the
+/// resource-value pattern.
+///
+/// `order_book` stores a `copy + drop` value; this vault stores `Coin<SUI>`, which has
+/// neither. That changes the lifecycle and is where integrators most often slip:
+///
+/// - A `SortedMap<u64, Coin<SUI>>` is itself not droppable - it cannot fall out of
+///   scope; it must be drained and then `destroy_empty`'d.
+/// - Every op that could displace a value hands it back instead of dropping it
+///   (`upsert` returns `Option<Coin>`, `remove!` returns `(rank, Coin)` (aborting if the rank
```
ericnordelo, line 10: The prize-vault example documents the wrong return type. It says remove! returns (rank, Coin), while the implemented map API returns only Coin; pay_rank correctly consumes only the coin.

collections/sources/sorted_set.move
```diff
@@ -0,0 +1,638 @@
...
+///
+/// A coarse (non-injective) comparator reports two byte-distinct keys as equal, collapsing them
+/// to one element. On that collision the last-written key's bytes win (upsert overwrites the stored
+/// key with the incoming one), so which byte-variant is retained is well-defined only under an
+/// injective comparator. In tests, call
+/// `sorted_map::is_well_formed_by!(inner(&set), lt)` after a `_by` sequence.
+///
+/// # `upsert` returns `bool`, not an abort
+///
+/// A deliberate divergence from `sui::vec_set`, whose `upsert` aborts on a duplicate. `upsert ->
```
ericnordelo: incorrectly calls the Sui operation upsert; it should refer to sui::vec_set::insert.

There is no upsert on sui::vec_set

ericnordelo (commented):
Looking good @0xNeshi , left another review, and after fixes we can merge it and leave it to auditors.

collections/sources/sorted_map.move
0xNeshi: Great catch, addressing

collections/examples/sorted_map/prize_vault.move
0xNeshi: This is referring to `SortedMap` ops. Updating docs to clarify that

ericnordelo (approved):
LGTM after fixing the CI issues.

bidzyyys:
## Review — openzeppelin_collections

- [x] ⚪ **sorted_map.move:419 (`from_sorted_keys_values_by!`)** — the map's bulk constructor never tests n==1 (singleton) or the first-pair abort, though the sibling `sorted_set::from_sorted_keys` tests both boundaries → add a `from_sorted_keys_values!(vector[1], vector[10])` singleton case and a `vector[2,1]` case asserting `EKeysNotStrictlyIncreasing` (`location = sm`) to the map suite.
- [x] ⚪ **sorted_set.move:268 (`from_sorted_keys_by!`)** — requires `$K: copy + drop`, but the set's own `from_keys_by!` needs only `drop` and the map's `from_sorted_keys_values_by!` needs neither; the extra `copy` comes solely from `*keys.borrow(i)` → either consume the vector (as the map does) to drop the `copy` bound, or note the stricter requirement in the doc.

PR #455, opened 2026-07-07
fix: correct same-tx CAS idiom in spend_vault docs and example

kosedogus:
A same-tx `allowance<T>()` read cannot serve as a CAS guard, the shared Vault is locked for the whole tx, so the read trivially matches and `EUnexpectedAllowance` never fires on a mismatch. 
- Docs now say to derive `expected` from an earlier (off-chain / prior-tx) read,
- `direct_delegation::change_budget` takes `expected: u64` from the caller instead of self-reading, 
- A new test pinning the stale-read abort. 
Library code unchanged (comment-only); 177/177 tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved allowance update safety: concurrent spends won’t be overwritten. When the provided expected value is stale (or the allowance entry is absent), the update now aborts instead of clobbering a newer spend.

* **Documentation**
  * Clarified `set_allowance` CAS “Lifecycle” behavior, including how same-transaction and cross-transaction check-then-act scenarios can fail safely.

* **Examples & Tests**
  * Updated the direct delegation budget-change example to pass an expected pre-read value.
  * Added a negative test asserting abort on stale expected values.


ericnordelo (approved):
LGTM!

bidzyyys (approved):
LGTM

bidzyyys:
## Review — allowance

- [x] ⚪ **spend_vault.move:98, 661-662** — new doc prose uses em dashes (`—`), which STYLEGUIDE.md:255 forbids in all prose/comments/docs (the file had none before this PR) → replace each `—` with a regular `-`.
- [x] ⚪ **examples/spend_vault/tests/direct_delegation_tests.move:210** — same em-dash violation in the new test's comment → replace `—` with `-`.

